PR #6, opened 2019-04-25
 Added func RestoreDeltaBlockBackupIncrementally() and modified test

shuo-wu:


deltablock.go
```diff
@@ -307,22 +308,166 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
...
+}
+
+func restoreBlockToFile(volumeName string, volDev *os.File, bsDriver BackupStoreDriver, blk BlockMapping) error {
+	blkFile := getBlockFilePath(volumeName, blk.BlockChecksum)
+	rc, err := bsDriver.Read(blkFile)
+	if err != nil {
+		return err
+	}
+	r, err := util.DecompressAndVerify(rc, blk.BlockChecksum)
+	rc.Close()
```
yasker: There was no `defer` before because it's not in a function. Now you should able to use `defer rc.Close()`. Notice you should put it one line above, after error handling for rc.

deltablock.go
```diff
@@ -307,22 +308,166 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
...
+	var volDev *os.File
+	if _, err := os.Stat(volDevName); os.IsNotExist(err) {
+		volDev, err = os.Create(volDevName)
+		if err != nil {
 			return err
 		}
-		if _, err := io.CopyN(volDev, r, DEFAULT_BLOCK_SIZE); err != nil {
+		logrus.Debugf("Created new file %v", volDevName)
+	} else {
+		volDev, err = os.OpenFile(volDevName, os.O_RDWR, 0600)
```
yasker: I am not sure about file mode 0600. Anything wrong with `os.Open` call?

deltablock.go
```diff
@@ -307,22 +308,166 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
...
+	// check time
+	lastBackupTime, err := time.Parse(time.RFC3339, lastBackup.CreatedTime)
+	if err != nil {
+		return fmt.Errorf("Failed to check lastBackup %v timestamp: %v ", lastBackup, err)
+	}
+	backupTime, err := time.Parse(time.RFC3339, backup.CreatedTime)
+	if err != nil {
+		return fmt.Errorf("Failed to check backup %v timestamp: %v ", backup, err)
+	}
+	if !backupTime.After(lastBackupTime) {
```
yasker: Typo. And we don't want to check timestamp since it's not technically required that you are only able to restore backup later than the previous one.

deltablock.go
```diff
@@ -307,22 +308,166 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
...
+
+	log.WithFields(logrus.Fields{
+		LogFieldReason:     LogReasonStart,
+		LogFieldEvent:      LogEventRestoreIncre,
+		LogFieldObject:     LogFieldSnapshot,
+		LogFieldSnapshot:   srcBackupName,
+		LogFieldOrigVolume: srcVolumeName,
+		LogFieldVolumeDev:  volDevName,
+		LogEventBackupURL:  backupURL,
+	}).Debugf("started Incrementally restoring from %v to %v", lastBackup, backup)
```
yasker: weird capitalized `I`

deltablock.go
```diff
@@ -307,22 +308,166 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
...
+
+	for b, l := 0, 0; b < len(backup.Blocks) || l < len(lastBackup.Blocks); {
+		if b >= len(backup.Blocks) {
+			lB := lastBackup.Blocks[l]
+			log.Debugf("Incrementally fill 0 for file %v: block %v in lastBackup %v.", volDevName, l, lastBackup.Name)
```
yasker: Remove the debug message here and below. Otherwise, you're going to flood the log even when the volume is only a few GB.

deltablock.go
```diff
@@ -336,6 +481,13 @@ func RestoreDeltaBlockBackup(backupURL, volDevName string) error {
 	return nil
 }
 
+func fillZeroBlockToFile(volDev *os.File, blk BlockMapping) error {
+	if _, err := volDev.WriteAt(make([]byte, DEFAULT_BLOCK_SIZE), blk.Offset); err != nil {
```
yasker: Prepare an empty block and reuse it. Otherwise, Go would need to generate an array then recycle it later very often.

test/backup_test.go
```diff
@@ -274,3 +277,150 @@ func (s *TestSuite) TestBackupBasic(c *C) {
...
+
+	// snap0: blk * 4
+	snap0, err := os.Create(volume.Snapshots[0].Name)
+	c.Assert(err, IsNil)
+	defer snap0.Close()
+	for i := int64(0); i < 4; i++ {
+		_, err := snap0.WriteAt(data, i*blockSize)
+		c.Assert(err, IsNil)
+	}
+	err = snap0.Truncate(volumeSize)
```
yasker: Truncate at the beginning or at the end. It's weird to see it in the middle of preparing the data.

deltablock.go
shuo-wu: `os.Open` implementation: `return OpenFile(name, O_RDONLY, 0)`

yasker: Seems you can use `os.Create` and don't worry if the file exists or not.

shuo-wu: Do you mean `os.OpenFile(volDevName, os.O_RDWR|os.O_CREATE, 0600)`?

yasker: https://golang.org/pkg/os/#Create

shuo-wu: Cannot use `os.Create` because it will somehow overwrite the existing file

PR #8, opened 2019-05-13
fixed bug for func RestoreDeltaBlockBackupIncrementally()

shuo-wu:
https://github.com/rancher/longhorn/issues/529


yasker:
https://github.com/rancher/longhorn/issues/529

PR #9, opened 2019-06-03
Asynchronus Incremental Backup

shreesha21:
The following code changes modify the CreateDeltaBackup function to perform the incremental backup asynchronusly instead of doing it as a blocking operation.

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -115,7 +115,7 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
 		return "", err
 	}
 	if delta.BlockSize != DEFAULT_BLOCK_SIZE {
-		return "", fmt.Errorf("Currently doesn't support different block sizes driver other than %v", DEFAULT_BLOCK_SIZE)
+		return "", fmt.Errorf("currently doesn't support different block sizes driver other than %v", DEFAULT_BLOCK_SIZE)
 	}
 	log.WithFields(logrus.Fields{
 		LogFieldReason:       LogReasonComplete,
@@ -138,6 +138,18 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
 		SnapshotName: snapshot.Name,
 		Blocks:       []BlockMapping{},
 	}
+
+	return performIncrementalBackup(config, delta, deltaBackup, lastBackup, bsDriver)
+}
+
+func performIncrementalBackup(config *DeltaBackupConfig, delta *Mappings, deltaBackup *Backup, lastBackup *Backup,
+	bsDriver BackupStoreDriver) (string, error) {
+
+	volume := config.Volume
+	snapshot := config.Snapshot
+	destURL := config.DestURL
+	deltaOps := config.DeltaOps
+
 	mCounts := len(delta.Mappings)
 	newBlocks := int64(0)
 	for m, d := range delta.Mappings {
@@ -203,7 +215,7 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
 		return "", err
 	}
 
-	volume, err = loadVolume(volume.Name, bsDriver)
+	volume, err := loadVolume(volume.Name, bsDriver)
 	if err != nil {
 		return "", err
 	}
--- a/deltablock.go
+++ b/deltablock.go
@@ -31,6 +31,7 @@ type DeltaBlockBackupOperations interface {
 	OpenSnapshot(id, volumeID string) error
 	ReadSnapshot(id, volumeID string, start int64, data []byte) error
 	CloseSnapshot(id, volumeID string) error
+	UpdateBackupStatus(id, volumeID string, backupProgress int, backupURL string, err string) error
 }
 
 const (
@@ -39,6 +40,9 @@ const (
 	BLOCKS_DIRECTORY      = "blocks"
 	BLOCK_SEPARATE_LAYER1 = 2
 	BLOCK_SEPARATE_LAYER2 = 4
+
+	PROGRESS_PERCENTAGE_BACKUP_SNAPSHOT = 95
+	PROGRESS_PERCENTAGE_BACKUP_TOTAL    = 100
 )
 
 func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
@@ -74,13 +78,13 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
 	if err := deltaOps.OpenSnapshot(snapshot.Name, volume.Name); err != nil {
 		return "", err
 	}
-	defer deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
 
 	var lastSnapshotName string
 	var lastBackup *Backup
 	if lastBackupName != "" {
 		lastBackup, err = loadBackup(lastBackupName, volume.Name, bsDriver)
 		if err != nil {
+			deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
 			return "", err
 		}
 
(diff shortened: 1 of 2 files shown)


deltablock.go
```diff
@@ -31,14 +31,19 @@ type DeltaBlockBackupOperations interface {
 	OpenSnapshot(id, volumeID string) error
 	ReadSnapshot(id, volumeID string, start int64, data []byte) error
 	CloseSnapshot(id, volumeID string) error
+	UpdateBackupID(id, volumeID string, backupName string) error
```
yasker: BackupID would be returned for the function, so it's not needed.

deltablock.go
```diff
@@ -31,14 +31,19 @@ type DeltaBlockBackupOperations interface {
...
+	UpdateBackupProgress(id, volumeID string, backupProgress int32, err string) error
```
yasker: We can make this `UpdateBackupStatus` and cover the backup URL as well.

Noted that BackupURL can be generated early in the function, as soon as we have the backup ID. So what we really use BackupURL for is only to mark this backup as ready.

And `int` is good enough.

deltablock.go
```diff
@@ -31,14 +31,19 @@ type DeltaBlockBackupOperations interface {
...
+	UpdateBackupURL(id, volumeID string, backupURL string) error
```
yasker: Merge with `UpdateBackupStatus`.

deltablock.go
```diff
@@ -138,85 +142,102 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
...
-		}
-		block := make([]byte, DEFAULT_BLOCK_SIZE)
-		blkCounts := d.Size / delta.BlockSize
-		for i := int64(0); i < blkCounts; i++ {
-			offset := d.Offset + i*delta.BlockSize
-			log.Debugf("Backup for %v: segment %v/%v, blocks %v/%v", snapshot.Name, m+1, mCounts, i+1, blkCounts)
-			err := deltaOps.ReadSnapshot(snapshot.Name, volume.Name, offset, block)
-			if err != nil {
-				return "", err
+	var progress float64
```
yasker: Just make it `int`. Do the conversion during the calculation instead of in every function call.

deltablock.go
```diff
@@ -138,85 +142,102 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
...
+	go func() {
+		deltaOps.UpdateBackupID(snapshot.Name, volume.Name, deltaBackup.Name)
+		defer deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
+		mCounts := len(delta.Mappings)
```
yasker: We should make this part two commits. First one is moving the backup logic here into a function without modification, and then the second one will convert it into a goroutine. The reason is this block of code's indentation was changed at the same time as the functional changed, make it hard to see what's the functional change and what's the indentation change. Make it two commits would show the difference clearly. Besides, we normally prefer the code inside goroutine to be small, since sematically it's likely a separate function otherwise it won't be easily run inside a goroutine.

test/backup_test.go
```diff
@@ -223,11 +243,20 @@ func (s *TestSuite) TestBackupBasic(c *C) {
 				"RandomKey":    "RandomValue",
 			},
 		}
-		backup, err := backupstore.CreateDeltaBlockBackup(config)
+		_, err := backupstore.CreateDeltaBlockBackup(config)
+		c.Assert(err, IsNil)
+
+		for {
```
yasker: Don't do an infinite loop. Use RetryInterval and RetryCounts instead.

Same below.

shreesha21:
All the comments are addressed in the PR:  Updating backup progress in DeltaBackup #10

deltablock.go
```diff
@@ -139,7 +142,15 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
...
 	}
 
-	return performIncrementalBackup(config, delta, deltaBackup, lastBackup, bsDriver)
+	go func() {
+		if backup, err := performIncrementalBackup(config, delta, deltaBackup, lastBackup, bsDriver); err != nil {
+			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, "", err.Error())
+		} else {
+			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, backup, "")
+		}
+		defer deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
```
yasker: It's already at the end of function, no need for `defer`.

Or move this line to immediately after `go func(){`

test/backup_test.go
```diff
@@ -175,6 +187,26 @@ func (s *TestSuite) getDestURL() string {
 	return "nfs://127.0.0.1:/opt/backupstore"
 }
 
+func (s *TestSuite) getBackupUtil(c *C, config *backupstore.DeltaBackupConfig, deltaOps *RawFileVolume) string {
```
yasker: Name the function `createAndWaitForBackup`

test/backup_test.go
```diff
@@ -175,6 +187,26 @@ func (s *TestSuite) getDestURL() string {
...
+	_, err := backupstore.CreateDeltaBlockBackup(config)
+	c.Assert(err, IsNil)
+
+	retryCount := 120
+	for j := 0; j < retryCount; j++ {
+		if len(deltaOps.BackupURL) > 0 {
+			break
+		}
+		if len(deltaOps.BackupError) > 0 {
```
yasker: assert backupError doesn't exist first, no need for `if`.

deltablock.go
```diff
@@ -139,22 +144,31 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
...
 func performIncrementalBackup(config *DeltaBackupConfig, delta *Mappings, deltaBackup *Backup, lastBackup *Backup,
-	bsDriver BackupStoreDriver) (string, error) {
+	bsDriver BackupStoreDriver) (int, string, error) {
 
 	volume := config.Volume
 	snapshot := config.Snapshot
 	destURL := config.DestURL
 	deltaOps := config.DeltaOps
 
+	var progress float64
```
yasker: Use int instead so you won't need to convert it every time.

test/backup_test.go
```diff
@@ -175,6 +189,36 @@ func (s *TestSuite) getDestURL() string {
...
 
+func (s *TestSuite) createAndWaitForBackup(c *C, config *backupstore.DeltaBackupConfig, deltaOps *RawFileVolume) string {
+	_, err := backupstore.CreateDeltaBlockBackup(config)
+	c.Assert(err, IsNil)
+
+	retryCount := 120
+	var bError, bURL string
+	for j := 0; j < retryCount; j++ {
+
+		deltaOps.lock.Lock()
```
yasker: You can add a function e.g. `GetBackupStatus` which returns the progress and url with the lock held.

deltablock.go
```diff
@@ -31,14 +31,17 @@ type DeltaBlockBackupOperations interface {
...
 const (
 	DEFAULT_BLOCK_SIZE = 2097152
 
-	BLOCKS_DIRECTORY      = "blocks"
-	BLOCK_SEPARATE_LAYER1 = 2
-	BLOCK_SEPARATE_LAYER2 = 4
+	BLOCKS_DIRECTORY                    = "blocks"
+	BLOCK_SEPARATE_LAYER1               = 2
+	BLOCK_SEPARATE_LAYER2               = 4
+	PROGRESS_PERCENTAGE_BACKUP_SNAPSHOT = 95
```
yasker: Add one extra blank line here. Below the lines are progress related.

PR #13, opened 2019-07-16
The volume.cfg still exists when volume deletion is incomplete

ywei88:
The volume.cfg now still exists when there is leftover in blocks in backupstore, as mentioned in https://github.com/longhorn/longhorn/issues/597

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -74,9 +74,19 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
 	}
 
 	volumeDir := getVolumePath(volumeName)
-	if err := driver.Remove(volumeDir); err != nil {
-		return err
+	volumeBlocksDirectory := getBlockPath(volumeName)
+	volumeBackupsDirectory := getBackupPath(volumeName)
+
+	if err := driver.Remove(volumeBackupsDirectory); err != nil {
+		return fmt.Errorf("failed to remove all the backups for volume %v", volumeName)
 	}
+	if err := driver.Remove(volumeBlocksDirectory); err != nil {
+		return fmt.Errorf("failed to remove all the blocks for volume %v", volumeName)
+	}
+	if err := driver.Remove(volumeDir); err != nil {
+		return fmt.Errorf("failed to remove backup volume %v directory in backupstore", volumeName)
+	}
+
 	log.Debug("Removed volume directory in backupstore: ", volumeDir)
 	log.Debug("Removed backupstore volume ", volumeName)
 
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/urfave/cli"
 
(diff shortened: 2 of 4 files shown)


backupstore.go
```diff
@@ -74,8 +74,16 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
 	}
 
 	volumeDir := getVolumePath(volumeName)
-	if err := driver.Remove(volumeDir); err != nil {
-		return err
+	volumeBlocksDirectory := getBlockPath(volumeName)
```
yasker: You need to remove all the backups first. Otherwise, some backups may refer to the non-exist blocks if the block cleanup failed.

cmd/remove.go
```diff
@@ -37,3 +38,43 @@ func doBackupRemove(c *cli.Context) error {
 	}
 	return nil
 }
+
+func BackupVolumeRemoveCmd() cli.Command {
+	return cli.Command{
+		Name:    "removebackupvolume",
```
yasker: Update the `rm` command to `backup rm --volume <volume_name>`

backupstore.go
```diff
@@ -74,8 +74,19 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
...
+	volumeBackupsDirectory := getBackupPath(volumeName)
+
+	if err := driver.Remove(volumeBackupsDirectory); err != nil {
+		return fmt.Errorf("failed to remove all the backups for volume %v", volumeName)
+	}
+	if err := driver.Remove(volumeBlocksDirectory); err != nil {
+		return fmt.Errorf("failed to remove all the blocks for volume %v", volumeName)
+	}
+	if !volumeBlocksExists(volumeName, driver) {
```
yasker: If the previous line completed, when can this be true?

cmd/remove.go
```diff
@@ -24,16 +31,26 @@ func cmdBackupRemove(c *cli.Context) {
 
 func doBackupRemove(c *cli.Context) error {
 	if c.NArg() == 0 {
-		return RequiredMissingError("backup URL")
+		return RequiredMissingError("dest URL")
```
yasker: It should be either args[0] as dest URL or the volume argument. You don't need to provide `destURL` to delete a backup volume.

backupstore.go
ywei88: It's just a double check, since I'm afraid if the blocks deletion is interrupted, there is no error returned for driver.remove function

ywei88: Actually in the previous logic, the directory deletion was interrupted but I didn't see any error returned

cmd/remove.go
ywei88: if there is no dest URL, how can I get the backupstore driver to remove volume?

yasker: oh, you're right. So we're reusing the argument as either backupURL or destURL for volume. That should be fine.

yasker: Need to document that in the usage though.

PR #14, opened 2019-07-29
Progress report for backup deletion

ywei88:


deltablock.go
```diff
@@ -625,10 +651,13 @@ func DeleteDeltaBlockBackup(backupURL string) error {
 		if err != nil {
 			return err
 		}
+		totalBlocks := len(backup.Blocks)
```
yasker: This is wrong...

There are multiple backups in the volumes, blocks for one backup cannot be used as the denominator.

You can use half of the progress for scanning all the backups, another half of the progress on deleting the blocks which are no longer used.

yasker:
Close this one.

PR #15, opened 2019-07-31
Add a deleting flag to backup

ywei88:


yasker:
Need engine and manager integration test case to make sure empty volume returns the empty list (rather than error out).

PR #16, opened 2019-07-31
Add a deleting flag to backup volume

ywei88:


deltablock.go
```diff
@@ -251,6 +251,7 @@ func performIncrementalBackup(config *DeltaBackupConfig, delta *Mappings, deltaB
 	volume.LastBackupName = backup.Name
 	volume.LastBackupAt = backup.SnapshotCreatedAt
 	volume.BlockCount = volume.BlockCount + newBlocks
+	volume.Deleting = false
```
yasker: No, we shouldn't change this because it can be the volume in deleting and the state shouldn't be reversed.

Also when a backup was about to start, we should make sure the volume is not `deleting`.

deltablock.go
```diff
@@ -83,7 +83,9 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, error) {
 	if err != nil {
 		return "", err
 	}
-
+	if volume.Deleting == true {
+		return "", fmt.Errorf("volume %v is being deleted, can't create", volume.Name)
```
yasker: `can't create a new backup`

PR #24, opened 2019-10-03
Avoid sending the remove request with empty objects

shuo-wu:
https://github.com/longhorn/longhorn/issues/774


yasker:
In fact it's for https://github.com/longhorn/longhorn/issues/768

PR #33, opened 2020-05-02
Ignore non existing previous backup.cfg files

joshimoo:
There are many reasons this could happen, for example:
- the user switched backup stores or deleted files manually
- another process deleted that backup after we pulled the volume.cfg
Therefore we should expect the absence of a prior backup and proceed
with a full backup instead.

longhorn/longhorn#1266


joshimoo:
Integration test completed successfully

PR #34, opened 2020-05-07
Fix backupstore incremental backup logic

joshimoo:
We always merged the blocks from the new and previous backup on the
backupstore if the previous backup was present even in the case of a non
incremental backup.

longhorn/longhorn#1294

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -32,6 +32,28 @@ type BlockMapping struct {
 	BlockChecksum string
 }
 
+type backupRequest struct {
+	lastBackup *Backup
+}
+
+func (r backupRequest) isIncrementalBackup() bool {
+	return r.lastBackup != nil
+}
+
+func (r backupRequest) getLastSnapshotName() string {
+	if r.lastBackup == nil {
+		return ""
+	}
+	return r.lastBackup.SnapshotName
+}
+
+func (r backupRequest) getBackupType() string {
+	if r.isIncrementalBackup() {
+		return "incremental"
+	}
+	return "full"
+}
+
 type DeltaBlockBackupOperations interface {
 	HasSnapshot(id, volumeID string) bool
 	CompareSnapshot(id, compareID, volumeID string) (*Mappings, error)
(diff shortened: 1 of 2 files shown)


joshimoo:
Will send a separate PR for the engine integration test and the updated go mod once this is merged.
I verified the fix manually.

deltablock.go
```diff
@@ -133,7 +133,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 
 	if lastSnapshotName != "" {
 		isIncrementalBackup = true
+		log.Debugf("performing incremental backup new snapshot %s old snapshot %s",
+			snapshot.Name, lastSnapshotName)
 	}
```
yasker, line 138: How about just an `else` here?

deltablock.go
```diff
@@ -133,7 +133,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 
 	if lastSnapshotName != "" {
 		isIncrementalBackup = true
```
yasker, line 135: Is this variable still usable by itself? Seems it equals to `lastBackup != nil`

deltablock.go
```diff
@@ -133,7 +133,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+		log.Debugf("performing incremental backup new snapshot %s old snapshot %s",
```
yasker, line 136: Debug format is not consistent with existing logs.

deltablock.go
```diff
@@ -133,7 +133,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+			snapshot.Name, lastSnapshotName)
 	}
+	if !isIncrementalBackup {
+		// In the case where we want to do a non incremental backup, we cannot use
```
yasker, line 140: Don't need to explain here.

deltablock.go
joshimoo: It's actually equal to lastSnapshotName != ""
I removed it from the performBackup function since it's no longer needed and it's more clear that if lastBackup == nil that it has to be a full backup.

I refactored the function a little in the new commit, to make the flow a little clearer.

deltablock.go
joshimoo: refactored the full method a little, to make the code flow more straight forward.

deltablock.go
joshimoo: The variable is primarily used for the early returns in err case.

deltablock.go
```diff
@@ -84,42 +84,40 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
 
-	var lastSnapshotName string
-	var lastBackup *Backup
 	isIncrementalBackup := false
-	if lastBackupName != "" {
-		lastBackup, err = loadBackup(lastBackupName, volume.Name, bsDriver)
+	backupType := "full"
+	lastSnapshotName := ""
+	var lastBackup *Backup
+	if lastBackupName := volume.LastBackupName; lastBackupName != "" {
```
yasker, line 95: This style is a bit weird...

Do this instead:
```
if volume.LastBackupName != "" {
    lastBackupName := volume.LastBackupName
...
}
```

deltablock.go
```diff
@@ -84,42 +84,40 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+		} else if backup.SnapshotName != "" && !deltaOps.HasSnapshot(backup.SnapshotName, volume.Name) {
 			log.WithFields(logrus.Fields{
 				LogFieldReason:   LogReasonFallback,
 				LogFieldObject:   LogObjectSnapshot,
 				LogFieldSnapshot: lastSnapshotName,
 				LogFieldVolume:   volume.Name,
-			}).Debug("Cannot find last snapshot in local storage, would process with full backup")
+			}).Debug("Cannot find last snapshot in local storage")
+		} else {
+			lastBackup = backup
```
yasker, line 117: Feel those four parameters can be a struct instead since no one is going to modify one of them. But not a must.

joshimoo: I agree, I pushed a new commit with a struct implementation. Wasn't sure if you liked the struct methods implementation so I can reset to the prior commit if you don't like the struct :)

deltablock.go
```diff
@@ -32,6 +32,25 @@ type BlockMapping struct {
 	BlockChecksum string
 }
 
+type backupRequest struct {
+	isIncrementalBackup bool
```
yasker, line 36: this is `lastBackup != nil` right? We can add another function.

yasker:
Let's go with the struct approach.

deltablock.go
```diff
@@ -32,6 +32,25 @@ type BlockMapping struct {
...
+	lastBackup          *Backup
+}
+
+func (r backupRequest) lastSnapshotName() string {
```
yasker, line 40: should be `getLastSnapshotName()`

deltablock.go
```diff
@@ -32,6 +32,25 @@ type BlockMapping struct {
...
+	if r.lastBackup == nil {
+		return ""
+	}
+	return r.lastBackup.SnapshotName
+}
+
+func (r backupRequest) backupType() string {
```
yasker, line 47: Should be `getBackupType()`

deltablock.go
joshimoo: Correct with the new control flow it is indeed equal to lastBackup != nil, since we only set last backup after verifying that we can do an incremental backup.

deltablock.go
```diff
@@ -84,42 +106,35 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+			}).Info("Cannot find previous backup in backupstore")
+		} else if backup.SnapshotName == snapshot.Name {
 			//Generate full snapshot if the snapshot has been backed up last time
-			lastSnapshotName = ""
-			log.Debug("Would create full snapshot metadata")
-		} else if lastSnapshotName != "" && !deltaOps.HasSnapshot(lastSnapshotName, volume.Name) {
-			// It's possible that the snapshot in backupstore doesn't exist
-			// in local storage
-			lastSnapshotName = ""
+			log.Debug("Create full snapshot metadata")
```
yasker, line 128: inconsistent log.

deltablock.go
```diff
@@ -300,6 +312,8 @@ func mergeSnapshotMap(deltaBackup, lastBackup *Backup) *Backup {
 		}
 	}
 
+	log.Debugf("mergeSnapshotMap new: %s last: %s", deltaBackup.Name,
```
yasker, line 315: inconsistent log.

PR #36, opened 2020-05-13
Fix SSL certificate check

joshimoo:
longhorn/longhorn#704


joshimoo:
Will update the certificate with a new one that has a longer duration, so we don't have random integration test failures a year from now :)

joshimoo:
Will create a new PR for the updated Certificates.

PR #37, opened 2020-05-22
Reduce concurrency issues for backup operations

joshimoo:
When starting a backup after checking all preconditions we create a
backup metadata file that only contains the backup name as well as an
empty creation time which we use as an in progress marker. We currently
do not return these backups to the ui.

For nfs we now always create a tmp file with the current nano timestamp
appended so that 2 backup operations can never operate on the same tmp
file. We also do not remove these explicitly instead we rely on the
rename sys call to flip over the tmp file to the desired destination
file.

We now also also have a check to see if there have been any changes on
the known backups since we started the garbage collection process.
This will lead to skipping the block deletion if we detect changes on
the backup count.

I also removed the Volume deletion as part of the Backup deletion call.
Since we have a dedicated call for volume deletion, we should only ever
delete the whole volume if it actually is requested by the caller via
the dedicated volume deletion call.

longhorn/longhorn#1341
longhorn/longhorn#1326

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -198,6 +198,11 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 func performBackup(config *DeltaBackupConfig, delta *Mappings, deltaBackup *Backup, lastBackup *Backup,
 	bsDriver BackupStoreDriver) (int, string, error) {
 
+	// create an in progress backup metadata file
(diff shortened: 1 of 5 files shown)


deltablock.go
```diff
@@ -2,12 +2,11 @@ package backupstore
 
 import (
 	"fmt"
+	"github.com/sirupsen/logrus"
```
yasker, line 5: Don't move this line.

joshimoo:
Waiting for QA test with prebuilt images.

PR #39, opened 2020-05-27
Fix backupstore list issues

joshimoo:
Previously when loading the backup names error we would return an error for the whole list call, instead, we should return volume info with an error set.

This way an issue on one volume cannot block the whole backup list from showing up.

longhorn/longhorn#1410


yasker:
@joshimoo We will need an engine test case and e2e test case for this. Can you file an issue? We might want to merge it for v1.0.1 instead.

PR #40, opened 2020-05-29
Fix backup block deletion

joshimoo:
We need to skip the backup to be deleted otherwise we would mark its
blocks as in use. Which would lead to no block deletion.

longhorn/longhorn#1431

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -631,26 +631,31 @@ func DeleteDeltaBlockBackup(backupURL string) error {
 	for _, blk := range backupToBeDeleted.Blocks {
 		discardBlockSet[blk.BlockChecksum] = true
 	}
-	discardBlockCounts := len(discardBlockSet)
 
 	log.Debug("GC started")
 	backupNames, err := getBackupNamesForVolume(volumeName, bsDriver)
-	for _, backupName := range backupNames {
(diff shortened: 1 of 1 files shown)


deltablock.go
```diff
@@ -631,27 +631,32 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
+		backup, err := loadBackup(backupNames[i], volumeName, bsDriver)
 		if err != nil {
 			return err
 		}
-		for _, blk := range backup.Blocks {
+
+		if isBackupInProgress(backup) {
+			log.Infof("skipping block deletion because of in progress backup %v for volume %v",
+				backup.Name, volumeName)
+			discardBlockSet = nil
```
yasker, line 646: Use `map[string]bool{}` instead of nil or `make(map[string]bool)` as when we defined it. Don't want to trigger NPE by accident. (it doesn't look like it will, but wants to be cautious at this time)

joshimoo: Will change it, but as far as I understand a nil map is equivalent to a empty map with the only difference being that you cannot assign values to an uninitialized map which would trigger an Exception.

yasker: You're right.

https://play.golang.org/p/zlMIviP6Iqc

I will leave it up to you then.

yasker:
LGTM.

PR #45, opened 2020-07-02
Backupstore file locks

joshimoo:
### Proposal
The idea is to implement a locking mechanism that utilizes the backupstore, 
to prevent the following dangerous cases of concurrent operations.
1. prevent backup deletion during backup restoration
2. prevent backup deletion while a backup is in progress
3. prevent backup creation during backup deletion
4. prevent backup restoration during backup deletion

The locking solution shouldn't unnecessary block operations, so the following cases should be allowed.
1. allow backup creation during restoration
2. allow backup restoration during creation

The locking solution should have a maximum wait time for lock acquisition, 
which will fail the backup operation so that the user does not have to wait forever.

The locking solution should be self expiring, so that when a process dies unexpectedly, 
future processes are able to acquire the lock.

The locking solution should guarantee that only a single type of lock is active at a time.

The locking solution should allow a lock to be passed down into async running go routines.

### Implementation Overview
Conceptually the lock can be thought of as a **RW** lock, 
it includes a `Type` specifier where different types are mutually exclusive.

To allow the lock to be passed into async running go routines, we add a `count` field,
that keeps track of the current references to this lock. 

```go
type FileLock struct {
	Name         string
	Type         LockType
	Acquired     bool
	driver       BackupStoreDriver
	volume       string
	count        int32
	serverTime   time.Time
	refreshTimer *time.Ticker
}
```

To make the lock self expiring, we rely on `serverTime` updates which needs to be refreshed by a timer.
We chose a `LOCK_REFRESH_INTERVAL` of **60** seconds, each refresh cycle a locks `serverTime` will be updated.
A lock is considered expired once the current time is after a locks `serverTime` + `LOCK_MAX_WAIT_TIME` of **150** seconds.
Once a lock is expired any currently active attempts to acquire that lock will timeout.

```go
const (
	LOCKS_DIRECTORY       = "locks"
	LOCK_PREFIX           = "lock"
	LOCK_SUFFIX           = ".lck"
	LOCK_REFRESH_INTERVAL = time.Second * 60
	LOCK_MAX_WAIT_TIME    = time.Second * 150
	LOCK_CHECK_INTERVAL   = time.Second * 10
	LOCK_CHECK_WAIT_TIME  = time.Second * 2
)
```

Lock Usage
1. create a new lock instance via `lock := lock.New()`
2. call `lock.Lock()` which will block till the lock has been acquired and increment the lock reference count.
3. defer `lock.Unlock()` which will decrement the lock reference count and remove the lock once unreferenced.

To make sure the locks are **mutually exclusive**, we use the following process to acquire a lock.
1. create a lock file on the backupstore with a unique `Name`.
2. retrieve all lock files from the backupstore order them by `Acquired` then by `serverTime` 
   followed by `Name`
3. check if we can acquire the lock, we can only acquire if there is no unexpired(i) lock 
   of a different type(ii) that has priority(iii).
   1. Locks are self expiring, once the current time is after 
      `lock.serverTime + LOCK_MAX_WAIT_TIME` we no longer need to consider 
      this lock as valid.
   2. Backup & Restore Locks are mapped to compatible types while Delete
      Locks are mapped to a different type to be mutually exclusive with the
      others.
   3. Priority is based on the comparison order, where locks are compared by
      `lock.Acquired` then by `lock.serverTime` followed by `lock.Name`. Where
      acquired locks are always sorted before non acquired locks.
4. if lock acquisition times out, return err which will fail the backup operation.
5. once the lock is acquired, continuously refresh the lock (updates `lock.serverTime`) 
5. once the lock is acquired, it can be passed around by calling `lock.Lock()`
6. once the lock is no longer referenced, it will be removed from the backupstore.

It's very unlikely to run into lock collisions, since we use uniquely generated name for the lock filename.
In cases where two locks have the same `lock.serverTime`, we can rely on the `lock.Name` as a differentiator between 2 locks.

### DIFF
--- a/lock.go
+++ b/lock.go
@@ -1,0 +1,251 @@
+package backupstore
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/longhorn/backupstore/util"
+)
+
+const (
+	LOCKS_DIRECTORY       = "locks"
+	LOCK_PREFIX           = "lock"
+	LOCK_SUFFIX           = ".lck"
+	LOCK_DURATION         = time.Second * 150
+	LOCK_MAX_WAIT_TIME    = time.Second * 30
+	LOCK_REFRESH_INTERVAL = time.Second * 60
+	LOCK_CHECK_INTERVAL   = time.Second * 10
+	LOCK_CHECK_WAIT_TIME  = time.Second * 2
+)
+
+type LockType int
+
+const UNTYPED_LOCK LockType = 0
+const BACKUP_LOCK LockType = 1
+const RESTORE_LOCK LockType = 1
+const DELETION_LOCK LockType = 2
+
+type FileLock struct {
+	Name       string
+	Type       LockType
+	Acquired   bool
+	driver     BackupStoreDriver
+	volume     string
+	count      int32
+	serverTime time.Time
+	keepAlive  chan struct{}
+}
+
+func New(driver BackupStoreDriver, volumeName string, lockType LockType) (*FileLock, error) {
+	return &FileLock{driver: driver, volume: volumeName, Type: lockType}, nil
+}
+
+// isExpired checks whether the current lock is expired
+func (lock *FileLock) isExpired() bool {
+	isExpired := time.Now().Sub(lock.serverTime) > LOCK_DURATION
+	return isExpired
+}
+
+func (lock *FileLock) canAcquire() bool {
+	canAcquire := true
+	locks := getLocksForVolume(lock.volume, lock.driver)
+	for _, serverLock := range locks {
+		serverLockHasDifferentType := serverLock.Type != lock.Type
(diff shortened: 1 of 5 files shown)


yasker:
LGTM. @shuo-wu can you take a look as well?

lock.go
```diff
@@ -0,0 +1,255 @@
...
+	for acquired := lock.Acquired; !acquired; acquired = lock.Acquired {
+		if lock.canAcquire() {
+			file := getLockFilePath(lock.volume, lock.Name)
+			log.Debugf("Acquired lock %v on backupstore", file)
+			lock.Acquired = true
+			atomic.AddInt32(&lock.count, 1)
+			break
+		}
+
+		if lock.isExpired() {
```
shuo-wu, line 110: It's better to do cleanup if the lock can not be acquired.

joshimoo: agree, fixed

PR #46, opened 2020-07-05
deltablock: Move the deletion of backup.cfg file to before GC process

PhanLe1010:
The reason that it takes considerable amount of time to delete the backup.cfg in back end and UI continues to show the backup is because [the deletion of backup.cfg file](https://github.com/longhorn/backupstore/blob/2458fbe4269704cbe0351b63c3bf83f3be8e2f81/deltablock.go#L726) is happening after the function call [getBlockInfoMap()](https://github.com/longhorn/backupstore/blob/2458fbe4269704cbe0351b63c3bf83f3be8e2f81/deltablock.go#L719). Function `getBlockInfoMap()` takes a long time because it is reading every block from from datastore. As the result, we experienced the above delay.

This fix move the deletion of backup.cfg file to be before the GC process (so it is also before the function call `getBlockInfoMap()`). This makes sure that Garbage Collector process does not blocking the UI from responding to the user. Also, because the backup.cfg is deleted immediately, we don't have the problem of having more than the number of retain count backups in recurring backup job

longhorn/longhorn#1471
longhorn/longhorn#1472

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -688,6 +688,26 @@ func DeleteDeltaBlockBackup(backupURL string) error {
 		}
 	}
 
+	// we can delete the requested backupToBeDeleted immediately before GC starts
+	if err := removeBackup(backupToBeDeleted, bsDriver); err != nil {
+		return err
+	}
+	log.Infof("Removed backup %v for volume %v", backupName, volumeName)
+
+	// update the volume
+	v, err := loadVolume(volumeName, bsDriver)
+	if err != nil {
+		return fmt.Errorf("Cannot find volume %v in backupstore due to: %v", volumeName, err)
+	}
+
+	if backupToBeDeleted.Name == v.LastBackupName {
+		v.LastBackupName = ""
+		v.LastBackupAt = ""
+		if err := saveVolume(v, bsDriver); err != nil {
+			return err
+		}
+	}
+
 	log.Debug("GC started")
 	var backupsToBeRetained []*Backup
 	deleteBlocks := true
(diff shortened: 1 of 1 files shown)


deltablock.go
```diff
@@ -723,12 +729,6 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
 
-	// we can delete the requested backupToBeDeleted now
-	if err := removeBackup(backupToBeDeleted, bsDriver); err != nil {
-		return err
-	}
-	log.Infof("Removed backup %v for volume %v", backupName, volumeName)
-
 	// update the volume
 	v, err := loadVolume(volumeName, bsDriver)
 	if err != nil {
```
joshimoo, line 734: Move the volume update code directly after the backup deletion code.

deltablock.go
```diff
@@ -688,6 +688,12 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
 
+	// we can delete the requested backupToBeDeleted immediately before GC starts
+	if err := removeBackup(backupToBeDeleted, bsDriver); err != nil {
+		return err
+	}
+	log.Infof("Removed backup %v for volume %v", backupName, volumeName)
+
 	log.Debug("GC started")
 	var backupsToBeRetained []*Backup
 	deleteBlocks := true
```
joshimoo, line 699: Remove the util.Filter call since the backup has now already been deleted.

yasker:
@joshimoo Can you review this? I am not sure if the order change will introduce more race conditions.

deltablock.go
PhanLe1010: Thanks! Removed

deltablock.go
joshimoo: @PhanLe1010 did you see this comment?

PhanLe1010: @joshimoo Sorry, I missed this one. Will fix it now

PhanLe1010: Is it a styling fix or a potential bug? I don't quite see the problem

joshimoo: Think of it this way, you would want to have the 2 actions that are dependent on each other, as close together as possible.
Also this will make it easier to do the block cleanup asynchronous since all the code that needs to be synchronously executed is close together :)

PhanLe1010: Thank you for the explanation!

PR #47, opened 2020-07-07
Check for NFSv3 on error

ttpcodes:
This PR adds a check if the `mount` fails to see if the `NFS` share is able to be mounted using `NFSv3`. If the share can be mounted, we can pass this information to the user so they can be aware that the `NFS` server may only support `NFSv3` and to fix it if that is the case.

Since the usual methods for querying `NFS` depend on `RPC`, we instead check for `NFSv3` by attempting to `mount` the failed share with `NFSv3`. If this `mount` succeeds, we clean up the `mount` and then tell the user that the server may only support `NFSv3`.

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -10,6 +10,7 @@ import (
 	"github.com/longhorn/backupstore"
 	"github.com/longhorn/backupstore/fsops"
 	"github.com/longhorn/backupstore/util"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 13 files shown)


nfs/nfs.go
```diff
@@ -79,8 +79,19 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 	return b, nil
 }
 
-func (b *BackupStoreDriver) mount() error {
-	var err error
+func (b *BackupStoreDriver) mount() (err error) {
+	defer func() {
+		if err != nil {
+			if _, err := util.Execute("mount", []string{"-t", "nfs", "-o", "nfsvers=3", "-o", "nolock", b.serverPath, b.mountDir}); err != nil {
```
yasker, line 85: overriding `err`?

We can use `errors.Wrapf()`

yasker:
Can we separate the vendor change from code change?

ttpcodes:
The commit has been split into the `vendor` commit and the `NFS` commit.

nfs/nfs.go
```diff
@@ -79,8 +80,19 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
+				return
```
yasker, line 87: If we just return here, will we return with the error message of the previous v4 mount or this v3 mount? If the latter, it might be confusing.

ttpcodes: This returns with the `error` message from the previous `NFSv4` mount. It only continues and wraps the error with the `NFSv3` error if there was no error here.

PR #51, opened 2020-07-13
Support virtual-hosted-style access for S3 compatible storage

shuo-wu:
longhorn/longhorn#1552

### DIFF
--- a/s3/s3_service.go
+++ b/s3/s3_service.go
@@ -18,13 +18,27 @@ type Service struct {
 	Client *http.Client
 }
 
+const (
+	VirtualHostedStyle = "VIRTUAL_HOSTED_STYLE"
+)
+
 func (s *Service) New() (*s3.S3, error) {
 	// get custom endpoint
 	endpoints := os.Getenv("AWS_ENDPOINTS")
 	config := &aws.Config{Region: &s.Region}
+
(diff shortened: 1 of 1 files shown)


s3/s3_service.go
```diff
@@ -18,13 +18,27 @@ type Service struct {
...
+)
+
 func (s *Service) New() (*s3.S3, error) {
 	// get custom endpoint
 	endpoints := os.Getenv("AWS_ENDPOINTS")
 	config := &aws.Config{Region: &s.Region}
+
+	virtualHostedStyleEnabled := os.Getenv(VirtualHostedStyle)
+	if virtualHostedStyleEnabled == "true" {
+		config.S3ForcePathStyle = aws.Bool(false)
```
yasker, line 32: Can we just always set `S3ForcePathStyle = false`?

yasker: In fact, it's riskier to always set it to false. Do we know when should we set the option to true?

shuo-wu: 1. `Set this to `true` to force the request to use path-style addressing`
2. No, we don't know what the actual backup target server is. For MinIO, the default access style is path-style. It's too aggressive to set the default value to false.

PR #52, opened 2020-07-13
add drone build

meldafrawi:
longhorn/longhorn#1560

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@rancher.com>

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -4,7 +4,15 @@ set -e
 NFS_SERVER=backupstore_nfs_test_server
 NFS_PATH="/opt/backupstore"
 
+docker ps -a | grep "${NFS_SERVER}"
+
+if [[ $? -eq 0 ]]; then
+	docker rm -f "${NFS_SERVER}"
+fi
+
 function cleanup {
+        echo Unmounting NFS Share
+        umount `mount -l | grep nfs | grep "${NFS_IPADDR}" |  awk '{print $3}'`
         echo Stopping NFS server
         docker rm -vf $NFS_SERVER
 }
@@ -14,7 +22,10 @@ cd $(dirname $0)/..
 
 echo Starting NFS server
 
-docker run -d --name $NFS_SERVER --net container:$(hostname) --privileged docker.io/erezhorev/dockerized_nfs_server $NFS_PATH
+docker pull longhornio/nfs-ganesha:latest
+docker run -d --name $NFS_SERVER --privileged longhornio/nfs-ganesha:latest
+
+export NFS_IPADDR=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ${NFS_SERVER}`
 
 echo Wait for NFS server\'s 90 seconds grace period to pass
 sleep 90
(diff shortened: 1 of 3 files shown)


yasker:
I guess we don't have `nfs-kernel-server` running on the drone node... Might need to try ganesha to remove the dependency.

scripts/test
```diff
@@ -14,7 +16,9 @@ cd $(dirname $0)/..
 
 echo Starting NFS server
 
-docker run -d --name $NFS_SERVER --net container:$(hostname) --privileged docker.io/erezhorev/dockerized_nfs_server $NFS_PATH
+docker run -d --name $NFS_SERVER --privileged meldafrawi/ganesha-nfs:latest
+
+export NFS_IPADDR=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ${NFS_SERVER}`
 
 echo Wait for NFS server\'s 90 seconds grace period to pass
```
yasker, line 23: Do we still 90 need seconds now? I remember it can be disabled in Ganesha.

scripts/test
```diff
@@ -14,7 +16,9 @@ cd $(dirname $0)/..
 
 echo Starting NFS server
 
-docker run -d --name $NFS_SERVER --net container:$(hostname) --privileged docker.io/erezhorev/dockerized_nfs_server $NFS_PATH
+docker run -d --name $NFS_SERVER --privileged meldafrawi/ganesha-nfs:latest
```
yasker, line 19: Need to fork the repo and move it to Longhorn.

scripts/test
```diff
@@ -5,6 +5,8 @@ NFS_SERVER=backupstore_nfs_test_server
 NFS_PATH="/opt/backupstore"
 
 function cleanup {
+        echo Unmounting NFS Share
+		umount `mount -l | grep nfs | awk '{print $3}'`
```
yasker, line 9: If run outside of the container, this will have a side affect of mounting NFS on the host.

We can add IP to make sure we're unmounting the right NFS.

test/backup_test.go
```diff
@@ -223,10 +223,19 @@ func (s *TestSuite) SetUpSuite(c *C) {
 func (s *TestSuite) TearDownSuite(c *C) {
 	err := exec.Command("rm", "-rf", s.BasePath).Run()
 	c.Assert(err, IsNil)
+
```
yasker, line 226: ?

test/backup_test.go
```diff
@@ -223,10 +223,19 @@ func (s *TestSuite) SetUpSuite(c *C) {
...
 }
 
 func (s *TestSuite) getDestURL() string {
 	//return "vfs://" + s.BackupStorePath
```
yasker, line 230: Let's remove this line.

scripts/test
meldafrawi: yes, creating a disk will take sometime, we don't want to start running the tests too early.

scripts/test
meldafrawi: repo forked, and ownership transferred

test/backup_test.go
meldafrawi: fixed

test/backup_test.go
meldafrawi: fixed

scripts/test
meldafrawi: fixed

PR #53, opened 2020-07-17
Ignore errors during listing of backup volume names

joshimoo:
This addresses a case in which the backup volume deletion and backup
volume listing operation happen concurrently. It's safe to ignore these
errors since they are only ignored for backup volume listing and the
specific volume will just be ignored, since it's no longer present.

longhorn/longhorn#710
longhorn/longhorn#1604

### DIFF
--- a/config.go
+++ b/config.go
@@ -97,31 +97,29 @@ func getVolumeFilePath(volumeName string) string {
 	return filepath.Join(volumePath, volumeCfg)
 }
 
-func getVolumeNames(driver BackupStoreDriver) ([]string, error) {
+// getVolumeNames returns all volume names based on the folders on the backupstore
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -97,31 +97,24 @@ func getVolumeFilePath(volumeName string) string {
...
-		return names, nil
-	}
+	lv1Dirs, _ := driver.List(volumePathBase)
 	for _, lv1 := range lv1Dirs {
 		lv1Path := filepath.Join(volumePathBase, lv1)
-		lv2Dirs, err := driver.List(lv1Path)
-		if err != nil {
-			return nil, err
-		}
+		lv2Dirs, _ := driver.List(lv1Path)
```
yasker, line 107: Ignore the error can result in hard to debug issues. We can put those errors in log instead without error out.

PR #54, opened 2020-07-31
Add file based locks for concurrent backupstore operations

joshimoo:
We can only acquire a lock if there is no unexpired(1) lock of a different
type(2) that has priority(3).

1) Locks are self expiring, once the current time is after
`lock.serverTime + LOCK_MAX_WAIT_TIME` we no longer need to consider
this lock as valid.

2) Backup & Restore Locks are mapped to compatible types while Delete
Locks are mapped to a different type to be mutually exclusive with the
others.

3) Priority is based on the comparison order, where locks are compared by
`lock.Acquired` then by `lock.serverTime` followed by `lock.Name`. Where
acquired locks are always sorted before non acquired locks.

After 2 minutes, the lock acquisition will timeout and the backup
operation will pass through that error and fail.

longhorn/longhorn#612

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -116,6 +116,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 		return "", false, err
 	}
 
+	lock, err := New(bsDriver, volume.Name, BACKUP_LOCK)
+	if err != nil {
+		return "", false, err
+	}
+
+	defer lock.Unlock()
+	if err := lock.Lock(); err != nil {
+		return "", false, err
+	}
+
 	if err := addVolume(volume, bsDriver); err != nil {
 		return "", false, err
 	}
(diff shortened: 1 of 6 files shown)


joshimoo:
engine pr pointing to dev is here longhorn/longhorn-engine#512

lock.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return canAcquire
+}
+
+func (lock *FileLock) Lock() error {
+	lock.mutex.Lock()
+	defer lock.mutex.Unlock()
+	if lock.Acquired {
+		atomic.AddInt32(&lock.count, 1)
+		_ = saveLock(lock)
```
shuo-wu, line 75: I am not sure if we can directly ignore the error here.

joshimoo: What is your concern?
This error is save to ignore because the lock has already been acquired and it doesn't matter whether the write operation to the s3 store fails, since the previous acquisition is still recorded.

deltablock.go
```diff
@@ -116,6 +116,16 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 		return "", false, err
 	}
 
+	lock, err := New(bsDriver, volume.Name, BACKUP_LOCK)
+	if err != nil {
+		return "", false, err
+	}
+
+	defer lock.Unlock()
```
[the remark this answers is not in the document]
joshimoo: What do you mean?
Here we are not ignoring the error instead we are failing the backup operation in the case where we are unable to acquire the lock.

lock.go
shuo-wu: Make sense.

deltablock.go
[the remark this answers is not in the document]
shuo-wu: Oh, Sorry. The content was the following when I reviewed this PR.
```
	defer lock.Unlock()
        lock, err := New(bsDriver, volume.Name, BACKUP_LOCK)
	if err != nil {
		return "", false, err
	}
```

I am not sure if it was a GitHub bug or I just misread the code...

PR #56, opened 2020-09-15
Remove retry mechanism for lock acquisition

joshimoo:
**Use UTC for all lock comparisons:**
Previously time.Now() was used which returns the time in local time and
therefore the lock expiration check would not be correct, since the
servers modification time is in UTC for S3 and potentially in a different
timezone for the nfs server.

**Remove retry mechanism for lock acquisition:**
we only try to acquire once, since backup operations generally take some
time there is no point in trying to wait for lock acquisition, better to
throw an error and let the calling code retry with an exponential backoff

Since this code is blocking, this potentially also prevents the system
from being unresponsive for 60s in the case where the developer incorrectly
called the blocking code assuming it would be a quick operation.

longhorn/longhorn#612

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -18,7 +18,7 @@ type BackupStoreDriver interface {
 	GetURL() string
 	FileExists(filePath string) bool
 	FileSize(filePath string) int64
-	FileTime(filePath string) time.Time
+	FileTime(filePath string) time.Time     // Needs to be returned in UTC
 	Remove(path string) error               // Bahavior like "rm -rf"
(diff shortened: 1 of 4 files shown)


joshimoo:
Running the full e2e test suite at the moment.
Just the core test passed already.

shuo-wu (approved):

boknowswiki (approved):
LGTM

joshimoo:
Thanks for reviewing, I will merge this myself once I fixed an outstanding test issue.

PR #60, opened 2021-01-15
Store and present backing image info for backup volumes

shuo-wu:
longhorn/longhorn#2006

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -8,12 +8,14 @@ import (
 )
 
 type Volume struct {
-	Name           string
-	Size           int64 `json:",string"`
-	CreatedTime    string
-	LastBackupName string
-	LastBackupAt   string
-	BlockCount     int64 `json:",string"`
+	Name             string
+	Size             int64 `json:",string"`
+	CreatedTime      string
+	LastBackupName   string
+	LastBackupAt     string
+	BlockCount       int64  `json:",string"`
+	BackingImageName string `json:",string"`
+	BackingImageURL  string `json:",string"`
 }
 
 type Snapshot struct {
--- a/list.go
+++ b/list.go
@@ -19,6 +19,9 @@ type VolumeInfo struct {
 	Messages map[MessageType]string
 
 	Backups map[string]*BackupInfo `json:",omitempty"`
+
+	BackingImageName string
+	BackingImageURL  string
 }
 
 type BackupInfo struct {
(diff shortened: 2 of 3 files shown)


backupstore.go
```diff
@@ -8,12 +8,14 @@ import (
...
-	LastBackupAt   string
-	BlockCount     int64 `json:",string"`
+	Name             string
+	Size             int64 `json:",string"`
+	CreatedTime      string
+	LastBackupName   string
+	LastBackupAt     string
+	BackingImageName string `json:",string"`
+	BackingImageURL  string `json:",string"`
+	BlockCount       int64  `json:",string"`
```
yasker, line 18: Why changing BlockCount's position?

Also, do we need to record both name and URL? What if in the new environment the name doesn't match?

list.go
```diff
@@ -9,12 +9,14 @@ import (
...
-	LastBackupAt   string
-	DataStored     int64 `json:",string"`
+	Name             string
+	Size             int64 `json:",string"`
+	Created          string
+	LastBackupName   string
+	LastBackupAt     string
+	BackingImageName string
+	BackingImageURL  string
+	DataStored       int64 `json:",string"`
```
yasker, line 19: Why change the `DataStore`'s position?

list.go
```diff
@@ -114,14 +118,16 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
-		LastBackupAt:   volume.LastBackupAt,
-		DataStored:     int64(volume.BlockCount * DEFAULT_BLOCK_SIZE),
-		Messages:       make(map[MessageType]string),
-		Backups:        make(map[string]*BackupInfo),
+		Name:             volume.Name,
+		Size:             volume.Size,
+		Created:          volume.CreatedTime,
+		LastBackupName:   volume.LastBackupName,
+		LastBackupAt:     volume.LastBackupAt,
+		BackingImageName: volume.BackingImageName,
```
yasker, line 126: Can we keep the new fields at the end?

backupstore.go
shuo-wu: I will put new fields at the end. Similar below.
As for recording both name and URL, See the comment in the engine PR:
```
Name will be used if the user doesn't specify the backing image when restoring this backup.
URL is designed to inform users which backing image the backup volume was using. Then they can check if they need to create a new backing image object before the restore.
```

yasker (approved):

PR #63, opened 2021-02-19
Use concurrency goroutines on List API with bounded workers and queue length

jenting:
#### Proposed Changes
Since the volume backups are stored in double hashed folders.
For 1 volume backup, 1 + 1 + 1 = 3 s3 list command queries.
For 32 volume backups, the worst case is 1 + 32 + 32 = 65 s3 list command queries.

In this PR, we propose to use concurrency goroutines on List API with bounded workers and queue length to increase the performance.

In my testing (4 CPUs + 1-second delay from LH cluster to S3 MinIO) with 32 volume backups, the list API time be:
- w/o this PR: 2m53.696s
- w/ this PR: 0m11.150s

The Golang benchmark comparison:
- w/o this PR:
```
goos: linux
goarch: amd64
pkg: github.com/longhorn/backupstore
BenchmarkListAllVolumeOnly10ms32volumes
BenchmarkListAllVolumeOnly10ms32volumes-12     	       2	 958257131 ns/op	  206040 B/op	    2511 allocs/op
BenchmarkListAllVolumeOnly100ms32volumes
BenchmarkListAllVolumeOnly100ms32volumes-12    	       2	 951389031 ns/op	  206000 B/op	    2510 allocs/op
BenchmarkListAllVolumeOnly250ms32volumes
BenchmarkListAllVolumeOnly250ms32volumes-12    	       2	 951629719 ns/op	  206584 B/op	    2514 allocs/op
BenchmarkListAllVolumeOnly500ms32volumes
BenchmarkListAllVolumeOnly500ms32volumes-12    	       2	 950923296 ns/op	  205996 B/op	    2511 allocs/op
PASS
ok  	github.com/longhorn/backupstore	11.449s
```
- w/ this PR:
```
goos: linux
goarch: amd64
pkg: github.com/longhorn/backupstore
BenchmarkListAllVolumeOnly10ms32volumes
BenchmarkListAllVolumeOnly10ms32volumes-12     	      27	  42846050 ns/op	  364356 B/op	    4595 allocs/op
BenchmarkListAllVolumeOnly100ms32volumes
BenchmarkListAllVolumeOnly100ms32volumes-12    	      27	  42777124 ns/op	  359994 B/op	    4586 allocs/op
BenchmarkListAllVolumeOnly250ms32volumes
BenchmarkListAllVolumeOnly250ms32volumes-12    	      26	  42917962 ns/op	  358563 B/op	    4574 allocs/op
BenchmarkListAllVolumeOnly500ms32volumes
BenchmarkListAllVolumeOnly500ms32volumes-12    	      27	  42815242 ns/op	  362163 B/op	    4611 allocs/op
PASS
ok  	github.com/longhorn/backupstore	4.772s
```

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1955


config.go
```diff
@@ -101,24 +102,41 @@ func getVolumeFilePath(volumeName string) string {
...
+			for _, lv2 := range lv2Dirs {
+				go func(lv2 string) {
+					defer wg.Done()
+					lv2Path := filepath.Join(lv1Path, lv2)
+					volumeNames, err := driver.List(lv2Path)
+					if err != nil {
+						log.Warnf("failed to list volume names for path: %v reason: %v", lv2Path, err)
+						return
+					}
+					names = append(names, volumeNames...)
```
shuo-wu, line 133: Slice in Golang is not thread safe!

list.go
```diff
@@ -96,19 +100,40 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
-		volumeInfo, err := addListVolume(volumeName, driver, volumeOnly)
-		if err != nil {
-			return nil, err
-		}
-		resp[volumeName] = volumeInfo
+		go func(volumeName string) {
+			defer wg.Done()
+			volumeInfo, err := addListVolume(volumeName, driver, volumeOnly)
+			if err != nil {
+				errs = append(errs, err.Error())
```
shuo-wu, line 119: Same here. It's not thread safe.

list.go
```diff
@@ -96,19 +100,40 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
+	if len(errs) > 0 {
+		return nil, errors.New(strings.Join(errs, "\n"))
+	}
+
+	var resp = make(map[string]*VolumeInfo)
+	f := func(k, v interface{}) bool {
+		resp[k.(string)] = v.(*VolumeInfo)
+		return true
 	}
+	m.Range(f)
```
shuo-wu, line 136: Can we use an anonymous function instead?
```
m.Range(func(k, v interface{}) bool {
		resp[k.(string)] = v.(*VolumeInfo)
		return true
	})
```

list.go
```diff
@@ -96,19 +100,40 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
 	}
-	resp := make(map[string]*VolumeInfo)
 	volumeNames := []string{volumeName}
 	if volumeName == "" {
 		volumeNames = getVolumeNames(driver)
 	}
 
+	var (
+		wg   sync.WaitGroup
+		m    sync.Map // Use thread-safe sync.Map
```
shuo-wu, line 110: Rename this. e.g.,`volumeInfoMap`. A variable named `m` is confusing.

shuo-wu (changes requested):

joshimoo:
Good idea parallelizing the deep lookups, but instead of spawning an unbounded amount of go routines.
Consider implementing a fixed worker pool + queue approach.

Then we can pass a function to the workers + path param.
We can then use this for similar deep functions, `getBlockNamesForVolume`,
The `DeleteObjects` call can also benefit from parallel deletion.

jenting:
> Good idea parallelizing the deep lookups, but instead of spawning an unbounded amount of go routines.
> Consider implementing a fixed worker pool + queue approach.
> 
> Then we can pass a function to the workers + path param.
> We can then use this for similar deep functions, `getBlockNamesForVolume`,
> The `DeleteObjects` call can also benefit from parallel deletion.

Sounds good, I'll implement it and test the worker + queue pool performance.

jenting:
This is ready to review.

joshimoo:
@jenting Can you do a another benchmark run to see how the performance is with the fixed worker pool?

joshimoo (approved):
LGTM can probably be made a little bit more generic but we can do that in a future refactor.

joshimoo:
@shuo-wu any outstanding feedback? if not feel free to merge.

config.go
```diff
@@ -98,28 +103,81 @@ func getVolumeFilePath(volumeName string) string {
...
+			for i := range lv2Dirs {
+				lv2Paths[i] = filepath.Join(path, lv2Dirs[i])
+			}
+			return lv2Paths, nil
+		})
+		lv1Trackers = append(lv1Trackers, lv1Tracker)
+	}
+
+	for _, lv1Tracker := range lv1Trackers {
+		payload, err := lv1Tracker.Result()
```
shuo-wu, line 147: If a specific job somehow takes a long time to finish the job, will this line get blocked?

list.go
```diff
@@ -103,18 +108,48 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
+		})
+		trackers = append(trackers, tracker)
+	}
+
+	var (
+		resp = make(map[string]*VolumeInfo)
+		errs []string
+	)
+	for _, tracker := range trackers {
+		payload, err := tracker.Result()
```
shuo-wu, line 142: Similar question here

config.go
jenting: Yes, this line will be blocked until the job finished (or timeout by S3 API call).

If you worried there is no default timeout value to wait for the job to be finished, we could change the worker dispatcher function from `QueueFunc` to `QueueTimedFunc`.

shuo-wu: Yeah. It's unexpected but possible. Besides, it will affect the overall performance. I am thinking if we can:
1. Add timeout for the lv1Trackers
2. Use a buffered channel between `lv1Trackers` and this `for` loop. `lv1Trackers` always push the results to the buffer, while this `for` loop always get the latest result from the buffer then start a `lv2Tracker`.

jenting: If I understand correctly, the goal you want to achieve is that if one of the lv1 job done and then run related lv2 job. Rather than the current implementation that waits all the lv1 jobs done and then run all the lv2 job.

Am I understand correctly?

joshimoo: Yeah I thought about that too, but decided not to mention it since for now this implementation would be good enough.

In general it would have been nice to have a queue with n workers where we just pass in a `struct job { path string, finalDestination bool }
Then we don't need to deal with multi level trackers and instead just collect the elements where final destination == true.
The rest just keeps adding the new elements onto the job queue, till final destination == true.

Even better would be `struct job {path string, func op}`
Then as part of the op we would keep adding elements to the queue or collect the final elements.

If you want to implement that great if not, we can refactor this later.

jenting: Personally, I'd prefer to refactor this later once we have the further issue on the list API timeout issue.

The reason I did not add the timeout for the lv1Trackers and lv2Trackers is that I want the driver layer like S3 to control the timeout itself, I remember that the S3 default timeout value is 30 seconds.

shuo-wu: Agree. If we want to implement it and it's not complicated, we can do it later. Right now adding a timeout is the only necessary modification.

jenting: I added a 30 seconds timeout for the job queue. Please take a look, thank u all.

shuo-wu (approved):

PR #64, opened 2021-02-24
Add KubernetesStatus metadata to volume.cfg

jenting:
#### Proposed Changes
add PVC + namespace + Workload/Pod name in KubernetesStatus to volume.cfg.

I did not choose to get the **KuberneteStatus** metadata from the per-PVC `backup-backup-xxxx.cfg` is because I don't want to increase the S3 API list calls. We already have an issue with too many S3 API calls which causes the command timeout.

Therefore, I choose to add the **KubernetesStatus** to `volume.cfg` directly to make the `longhorn backup list --volume-only` to have the PVC + namespace + Workerload/Pod name information, then GUI could display this information in the backup list section of Dashboard.

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1539

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -10,6 +10,7 @@ import (
 type Volume struct {
 	Name             string
 	Size             int64 `json:",string"`
+	Labels           map[string]string
 	CreatedTime      string
 	LastBackupName   string
 	LastBackupAt     string
--- a/deltablock.go
+++ b/deltablock.go
@@ -325,6 +325,7 @@ func performBackup(config *DeltaBackupConfig, delta *Mappings, deltaBackup *Back
 	volume.BlockCount = volume.BlockCount + newBlocks
 	// The volume may be expanded
 	volume.Size = config.Volume.Size
+	volume.Labels = config.Labels
 	volume.BackingImageName = config.Volume.BackingImageName
 	volume.BackingImageURL = config.Volume.BackingImageURL
 
--- a/list.go
+++ b/list.go
@@ -2,6 +2,7 @@ package backupstore
 
 import (
 	"fmt"
+
 	"github.com/sirupsen/logrus"
 
 	. "github.com/longhorn/backupstore/logging"
@@ -11,6 +12,7 @@ import (
 type VolumeInfo struct {
 	Name           string
 	Size           int64 `json:",string"`
+	Labels         map[string]string
 	Created        string
 	LastBackupName string
 	LastBackupAt   string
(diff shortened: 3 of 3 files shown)


shuo-wu (approved):

yasker:
@joshimoo I remembered we decided to keep volume.cfg as static as possible to avoid the race conditions. Can you check if that applies here too?

jenting:
> @joshimoo I remembered we decided to keep volume.cfg as static as possible to avoid the race conditions. That's why we removed the `lastBackup` field from the volume.cfg. Can you check if that applies here too?

Not sure if I understand correctly. The `KubernetesStatus` contains this information right now
- PV name, PV status
- namespace
- PVC name
- last PVC reference at
- workload status:
  - pod name, pod status
  - workload name, workload type
- last Pod reference at


Is the `lastBackup` field you mention mean the last PVC/Pod reference at?

yasker:
@jenting I was referring to `volume.LastBackupName` and `volume.LastBackupAt`. It seems we didn't remove the fields themselves. If we still have the field, can we pull the information using `LastBackupName` instead of adding more fields to the volume? Since the last backup (the latest backup) should have the workload information and pulling it will only result in one extra API call.

jenting:
> @jenting I was referring to `volume.LastBackupName` and `volume.LastBackupAt`. It seems we didn't remove the fields themselves. If we still have the field, can we pull the information using `LastBackupName` instead of adding more fields to the volume? Since the last backup (the latest backup) should have the workload information and pulling it will only result in one extra API call.

Yes, it's one extra API call for one volume. However, since we haven't implemented the pagination on the backup page, so the extra API call relies on how many volume backups. For example, if we have 32 volume backups, currently API calls is 1 + 32 + 32 = 65. If we get the KubernetesStatus from the latest backup, the API calls become 65+32=97.

Or we could implement it by one extra API call once we have increased the performance (https://github.com/longhorn/longhorn/issues/1955) and implement the pagination (https://github.com/longhorn/longhorn/issues/2183) of the backup page. Then, we don't need extra metadata to saves to `volume.cfg`.

joshimoo:
I think this is safe to do now after my fixes + locks.

Just for information, there is still a potential race when updating the total volume block count, 
which is only used for visual information, so that shouldn't impact anything.
https://github.com/longhorn/backupstore/blob/e3b5c5a6d3ae70c1ca33d42aa1d882bc8eb5a051/deltablock.go#L325

condition: when 2 backups of the same volume complete at the same time. 
After they both have loaded the volume before the other has updated the volume.
The block count will be out of sync with reality but if any backup gets deleted, we always update the block count to the correct values.

---------------------------------------------

If we want to do something more complicated and lay the ground work for restoring the workload.
We can add a new workload.conf that has the serialized content from the deployment, pvc, pv yamls.
But for now this one is a simple implementation that we can change down the line.

---------------------------------------------

Conceptually the Kubernetes Status is only accurate at the time of the backup.
So it would be okay to use the Latest Backup, but that would make the implementation more complicated, so I think this is a more simpler approach.

joshimoo (approved):
LGTM, left a comment this is a simple implementation that should be safe now with all the concurrency fixes.

yasker (approved):
Thanks for the information @joshimoo . LGTM.

PR #66, opened 2021-03-09
Error handling backupstore.List when fsops path not exist

c3y1huang:
When the volume directory does not exist on NFS backstore should return empty list instead of error.

https://github.com/longhorn/longhorn/issues/2312


c3y1huang:
Found this only resolve first-level directory issue.

c3y1huang:
Running regression [Build #25](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/25/).

shuo-wu (approved):
LGTM

innobead (approved):

PR #67, opened 2021-03-10
Support AWS IAM Role

jenting:
#### Proposed Changes
Call `ses.Config.Credentials.Get()`, the AWS SDK would get the temporary role credentials from AWS AssumeRole.
Reference to https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1526

### DIFF
--- a/s3/s3_service.go
+++ b/s3/s3_service.go
@@ -49,6 +49,9 @@ func (s *Service) New() (*s3.S3, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM. 

Note: FYI. `aws-sdk` has some dynamic credential provisioner selection behind it, so it will try to dynamically pick the appropriate credential as per the environment.

PR #69, opened 2021-04-09
Fix delete backup memory spike

c3y1huang:
Each volume backup is most likely to reference the same block in the storage target. And having all backups with block metadata and volume block metadata to check reference count is memory-consuming.

Remove the use of the retained backup list and process a single backup metas at a time when doing the reference check.

https://github.com/longhorn/longhorn/issues/2339

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -710,67 +710,40 @@ func DeleteBackupVolume(volumeName string, destURL string) error {
 	return nil
 }
 
-func getBlockInfoMap(backups []*Backup, volume string, driver BackupStoreDriver) (map[string]*BlockInfo, error) {
-	blockInfos := make(map[string]*BlockInfo)
-	blockNames, err := getBlockNamesForVolume(volume, driver)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, name := range blockNames {
-		blockInfos[name] = &BlockInfo{
-			checksum: name,
-			path:     getBlockFilePath(volume, name),
-			refcount: 0,
+func checkBlockReferenceCount(blockInfos map[string]*BlockInfo, backup *Backup, volumeName string, driver BackupStoreDriver) {
+	for _, block := range backup.Blocks {
+		info, known := blockInfos[block.BlockChecksum]
+		if !known {
+			log.Errorf("backup %v refers to unknown block %v",
+				backup.Name, block.BlockChecksum)
+			info = &BlockInfo{checksum: block.BlockChecksum}
+			blockInfos[block.BlockChecksum] = info
 		}
+		info.refcount += 1
 	}
-
-	for _, backup := range backups {
-		for _, block := range backup.Blocks {
-			info, known := blockInfos[block.BlockChecksum]
-			if !known {
-				log.Errorf("backup %v of volume %v refers to unknown block %v",
-					backup.Name, volume, block.BlockChecksum)
-				info = &BlockInfo{checksum: block.BlockChecksum}
-				blockInfos[block.BlockChecksum] = info
-			}
-
-			info.refcount += 1
-		}
-	}
-
-	return blockInfos, nil
 }
 
-// This function will get the last backup from the remaining backups
-func GetLatestBackup(backupsToBeRetained []*Backup, vol *Volume) error {
-	var lastBackupName string
-	var lastBackupAt string
-
-	for _, backup := range backupsToBeRetained {
-		if lastBackupName == "" || lastBackupAt == "" {
-			lastBackupName = backup.Name
-			lastBackupAt = backup.SnapshotCreatedAt
-		}
-
-		backupTime, err := time.Parse(time.RFC3339, backup.SnapshotCreatedAt)
-		if err != nil {
-			return fmt.Errorf("Cannot parse backup %v time %v due to %v", backup.Name, backup.SnapshotCreatedAt, err)
-		}
-
-		lastBackupTime, err := time.Parse(time.RFC3339, lastBackupAt)
-		if err != nil {
-			return fmt.Errorf("Cannot parse  last backup %v time %v due to %v", lastBackupName, lastBackupAt, err)
-		}
-
-		if backupTime.After(lastBackupTime) {
-			lastBackupName = backup.Name
-			lastBackupAt = backup.SnapshotCreatedAt
-		}
+// getLatestBackup replace lastBackup object if the found
+// backup.SnapshotCreatedAt time is greater than the lastBackup
+func getLatestBackup(backup *Backup, lastBackup *Backup) error {
+	if lastBackup.SnapshotCreatedAt == "" {
+		*lastBackup = *backup
+		return nil
 	}
 
-	vol.LastBackupName = lastBackupName
-	vol.LastBackupAt = lastBackupAt
+	backupTime, err := time.Parse(time.RFC3339, backup.SnapshotCreatedAt)
+	if err != nil {
+		return fmt.Errorf("Cannot parse backup %v time %v due to %v", backup.Name, backup.SnapshotCreatedAt, err)
+	}
+
+	lastBackupTime, err := time.Parse(time.RFC3339, lastBackup.SnapshotCreatedAt)
+	if err != nil {
+		return fmt.Errorf("Cannot parse last backup %v time %v due to %v", lastBackup.Name, lastBackup.SnapshotCreatedAt, err)
+	}
+
+	if backupTime.After(lastBackupTime) {
+		*lastBackup = *backup
+	}
 
 	return nil
 }
@@ -785,12 +758,15 @@ func DeleteDeltaBlockBackup(backupURL string) error {
 	if err != nil {
 		return err
 	}
+	log := log.WithFields(logrus.Fields{
+		"backup": backupName,
+		"volume": volumeName,
+	})
 
 	lock, err := New(bsDriver, volumeName, DELETION_LOCK)
 	if err != nil {
 		return err
 	}
-
 	if err := lock.Lock(); err != nil {
 		return err
 	}
@@ -799,7 +775,7 @@ func DeleteDeltaBlockBackup(backupURL string) error {
 	// If we fail to load the backup we still want to proceed with the deletion of the backup file
 	backupToBeDeleted, err := loadBackup(backupName, volumeName, bsDriver)
 	if err != nil {
-		log.Warnf("failed to load to be deleted backup %v for volume %v", backupName, volumeName)
+		log.WithError(err).Warn("failed to load to be deleted backup")
 		backupToBeDeleted = &Backup{
 			Name:       backupName,
 			VolumeName: volumeName,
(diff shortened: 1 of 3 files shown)


c3y1huang:
```
=== RUN   Test
DEBU[0039] GC started                                    backup=backup-4051b2211c8d48de pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0039] deleted block c78b0b0afd0acf335ae3f97963afe9caecba2394df9bff06442a8a407d9ab009 for volume DeltaBlockTestVolume  backup=backup-4051b2211c8d48de pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0039] Retained 4 blocks for volume DeltaBlockTestVolume  backup=backup-4051b2211c8d48de pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0039] Removed 1 unused blocks for volume DeltaBlockTestVolume  backup=backup-4051b2211c8d48de pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0039] GC completed                                  backup=backup-4051b2211c8d48de pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0042] GC started                                    backup=backup-34a8c5dee6a24df0 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0042] deleted block 5f4b524e35962f33e85d95c1ae6121e4fbf5991b0fe770230da123f439d05f8e for volume DeltaBlockTestVolume  backup=backup-34a8c5dee6a24df0 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0042] Retained 3 blocks for volume DeltaBlockTestVolume  backup=backup-34a8c5dee6a24df0 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0042] Removed 1 unused blocks for volume DeltaBlockTestVolume  backup=backup-34a8c5dee6a24df0 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0042] GC completed                                  backup=backup-34a8c5dee6a24df0 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0044] GC started                                    backup=backup-f54a42a211794ae2 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0044] Retained 2 blocks for volume DeltaBlockTestVolume  backup=backup-f54a42a211794ae2 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0044] Removed 1 unused blocks for volume DeltaBlockTestVolume  backup=backup-f54a42a211794ae2 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0046] GC started                                    backup=backup-d124aad87f324f1b pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0046] deleted block ac501e6a6be0b20d11a28b6df5be2b41b4db05fdcf9b5b68cfe9603dffaec9e1 for volume DeltaBlockTestVolume  backup=backup-d124aad87f324f1b pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0046] Retained 1 blocks for volume DeltaBlockTestVolume  backup=backup-d124aad87f324f1b pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0046] Removed 1 unused blocks for volume DeltaBlockTestVolume  backup=backup-d124aad87f324f1b pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0046] GC completed                                  backup=backup-d124aad87f324f1b pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0048] GC started                                    backup=backup-8db07bf5553446a1 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0048] deleted block 224ad35246dbfac7d49c12d6ddef59fc6da3f5e46736fcef69ae12ab7779571d for volume DeltaBlockTestVolume  backup=backup-8db07bf5553446a1 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0048] Retained 0 blocks for volume DeltaBlockTestVolume  backup=backup-8db07bf5553446a1 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0048] Removed 1 unused blocks for volume DeltaBlockTestVolume  backup=backup-8db07bf5553446a1 pkg=backupstore volume=DeltaBlockTestVolume
DEBU[0048] GC completed                                  backup=backup-8db07bf5553446a1 pkg=backupstore volume=DeltaBlockTestVolume
OK: 1 passed
--- PASS: Test (48.62s)
PASS
ok      github.com/longhorn/backupstore/test    49.633s
```

deltablock.go
```diff
@@ -810,72 +786,86 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
+		v.LastBackupAt = ""
+	}
 
 	log.Debug("GC started")
-	var backupsToBeRetained []*Backup
 	deleteBlocks := true
 	backupNames, err := getBackupNamesForVolume(volumeName, bsDriver)
 	if err != nil {
-		log.Warnf("skipping block deletion failed to load backup names for volume %v", volumeName)
+		log.Warn("skipping block deletion, failed to load backup names")
```
shuo-wu, line 797: Need to print out the error message, and mention there will be a fallback then

deltablock.go
```diff
@@ -810,72 +786,86 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
+			checksum: name,
+			path:     getBlockFilePath(volumeName, name),
+			refcount: 0,
+		}
+	}
+
+	lastBackup := &Backup{}
 	for _, name := range backupNames {
+		log := log.WithField("backup", name)
 		backup, err := loadBackup(name, volumeName, bsDriver)
```
shuo-wu, line 817: Can you mention the issue with the GitHub link here?

shuo-wu (changes requested):
I don't remember why we add the lock outside of the goroutine then release the lock inside goroutine. Let us wait for Joshua to check it.

And I am not sure why you directly use `lock.mutex` rather than `lock.Lock()`... And I don't think it's the correct way to use this file lock.

deltablock.go
```diff
@@ -213,14 +213,10 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 		Blocks:       []BlockMapping{},
 	}
 
-	// keep lock alive for async go routine.
-	if err := lock.Lock(); err != nil {
-		deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
-		return "", backupRequest.isIncrementalBackup(), err
-	}
```
joshimoo, line 220: The reason we do this outside of the go routine, is that the lock can be thought of as a Semaphore (Counted Lock)
So we need to ensure that our Count of the Lock has been increased and persisted to the backupstore before the deferred Unlock gets called.

There is no guarantee that the GO routine has started before this function exists.

deltablock.go
```diff
@@ -213,14 +213,10 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
 	go func() {
+		lock.mutex.Lock()
+		defer lock.mutex.Unlock()
```
joshimoo, line 218: Doing the Unlock after the Close Snapshot is okay, but you need to use
`defer lock.Unlock()` so that you use the distributed lock. this applies to all the other replacements.

deltablock.go
joshimoo: This needs to be changed again for all the other replacements as well.

deltablock.go
```diff
@@ -710,67 +701,40 @@ func DeleteBackupVolume(volumeName string, destURL string) error {
...
-			lastBackupAt = backup.SnapshotCreatedAt
-		}
-
-		backupTime, err := time.Parse(time.RFC3339, backup.SnapshotCreatedAt)
-		if err != nil {
-			return fmt.Errorf("Cannot parse backup %v time %v due to %v", backup.Name, backup.SnapshotCreatedAt, err)
-		}
+// GetLatestBackup replace lastBackup object if the found
+// backup.SnapshotCreatedAt time is greater than the lastBackup
+func GetLatestBackup(backup *Backup, lastBackup *Backup) error {
```
joshimoo, line 719: Do we still need to export this function?

test/backup_test.go
```diff
@@ -14,13 +14,12 @@ import (
 	"testing"
 	"time"
 
-	//"github.com/sirupsen/logrus"
+	// "github.com/sirupsen/logrus"
```
joshimoo, line 17: If the logger is not used remove the comment here and below.

joshimoo (changes requested):
Your Lock usage is wrong, you replaced the distributed FileLock with an in memory mutex?
Have a look what the Lock/Unlock methods actually do.
https://github.com/longhorn/backupstore/blob/3f832934910ed93b3aed202ec63799b0db5af5e9/lock.go#L80

Good job on adding the tests, if you are having issues identifying the data race, let me know and I will try to help :)

Besides the lock replacement rest locks good.

c3y1huang:
Moved this back to draft to fix data race.

c3y1huang:
The data race seems to be coming from the logger. Removed the lock fix commit.

deltablock.go
```diff
@@ -810,72 +786,89 @@ func DeleteDeltaBlockBackup(backupURL string) error {
...
+			if err != nil {
+				log.WithError(err).Warn("failed to find last backup, skip block deletion")
+				deleteBlocks = false
+				break
+			}
 		}
+	}
+	if updateLastBackup {
+		v.LastBackupName = lastBackup.Name
+		v.LastBackupAt = lastBackup.SnapshotCreatedAt
```
joshimoo, line 855: This would potentially pick an arbitrary backup, depending on processing order (i.e. errors above)
- We either need to ensure that we pick the latest backup even if we skip block deletion. (which I think is risky)
-  Unset this, in the case where we skip block deletion, since that would lead to a full backup evaluation for the next backup. (I think this should also work for the DR volume, can you verify this)

Right now I tend to think the unset the latest backup in the case where we skip block deletion and have deleted the latest backup is the better approach, since it will lead to a full backup in the worst case, while arbitrarily picking one seems worse.

joshimoo (commented):
Left a comment about the latest backup selection, rest looks good to me.

joshimoo:
Leaving a note about the data race here:

That data race happened, because of an assignment to the global `log` var inside of the function.
Since this global is shared across all goroutines, it cannot be assigned to with out synchronization or a data race will happen, instead the assignment needs to happen to a new var, that can shadow the existing one. i.e. `log = X` needs to become `log := Y`
see https://github.com/longhorn/backupstore/blob/master/driver.go#L35

deltablock.go
shuo-wu: You mean preventing some operations like backup deletion happens after this function returns but before the go goroutine starts the actual backup creation?

deltablock.go
c3y1huang: I've mentioned the issue just above `checkBlockReferenceCount` since the fix is mostly to resolve the large array used previously during the reference count.

shuo-wu: You mean in the commit comment? There is no comment above function `checkBlockReferenceCount`

c3y1huang: https://github.com/longhorn/backupstore/pull/69/files#diff-507431733cf7ac7a656d8e1658e89dcfc4dc30f8c76e264f648c4ce36ee1c823R839-R842

deltablock.go
c3y1huang: > * Unset this, in the case where we skip block deletion, since that would lead to a full backup evaluation for the next backup. (I think this should also work for the DR volume, can you verify this)

[Add test_backup_block_deletion_error_then_dr_restore](https://github.com/longhorn/longhorn-tests/pull/592)

c3y1huang: Regression [https://ci.longhorn.io/job/private/job/longhorn-tests-regression/153/testReport/tests/test_basic/](Build#153).

deltablock.go
c3y1huang: Dropped lock fix commit.

deltablock.go
c3y1huang: No, thanks and fixed.

test/backup_test.go
c3y1huang: Addressed.

c3y1huang:
@joshimoo , @shuo-wu , I've addressed all review comments, PTAL.

joshimoo (approved):
LGTM, I think the DR test you posted needs to be updated though, to create the DR volume then wait for restore of backup(1) before deletion of backup(1). I am assuming the DR volume will stay on backup(1) till the next backup is created, which will update lastBackup.

joshimoo:
@shuo-wu any outstanding issues?

innobead (approved):
LGTM :+1:

deltablock.go
shuo-wu: Oh, sorry. My bad

shuo-wu (approved):
LGTM

PR #70, opened 2021-04-20
Fix XFS DR restore

c3y1huang:
~Replace io.CopyN so no extra data is copied.~ Pre-truncate volDev to ensure the XFS speculatively preallocates post-EOF blocks get reclaimed when the file is closed.

The issue is raised with the test case `test_dr_volume_with_backup_block_deletion`. The restored volume data contains blocks of messed up and zero bytes at the first backup restore ~, possibly due to the misalignment~ . During the investigation, it showed the block number written is 8064, almost twice the expected of 4096.

https://github.com/longhorn/longhorn/issues/2503

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -469,8 +469,21 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
 		defer volDev.Close()
 		defer lock.Unlock()
 
-		blkCounts := len(backup.Blocks)
 		var progress int
+		// This pre-truncate is to ensure the XFS speculatively
+		// preallocates post-EOF blocks get reclaimed when volDev is
+		// closed.
+		// https://github.com/longhorn/longhorn/issues/2503
+		// We want to truncate regular files, but not device
+		if stat.Mode()&os.ModeType == 0 {
+			log.Debugf("Truncate %v to size %v", volDevName, vol.Size)
+			if err := volDev.Truncate(vol.Size); err != nil {
+				deltaOps.UpdateRestoreStatus(volDevName, progress, err)
+				return
+			}
+		}
+
+		blkCounts := len(backup.Blocks)
 		for i, block := range backup.Blocks {
 			log.Debugf("Restore for %v: block %v, %v/%v", volDevName, block.BlockChecksum, i+1, blkCounts)
 			if err := restoreBlockToFile(srcVolumeName, volDev, bsDriver, block); err != nil {
@@ -481,14 +494,6 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
 			deltaOps.UpdateRestoreStatus(volDevName, progress, err)
 		}
 
-		// We want to truncate regular files, but not device
-		if stat.Mode()&os.ModeType == 0 {
-			log.Debugf("Truncate %v to size %v", volDevName, vol.Size)
-			if err := volDev.Truncate(vol.Size); err != nil {
-				deltaOps.UpdateRestoreStatus(volDevName, progress, err)
-				return
-			}
-		}
 		deltaOps.UpdateRestoreStatus(volDevName, PROGRESS_PERCENTAGE_BACKUP_TOTAL, nil)
 	}()
 
@@ -614,10 +619,10 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
 		defer volDev.Close()
 		defer lock.Unlock()
 
-		if err := performIncrementalRestore(srcVolumeName, volDev, lastBackup, backup, bsDriver, config); err != nil {
-			deltaOps.UpdateRestoreStatus(volDevName, 0, err)
-			return
-		}
+		// This pre-truncate is to ensure the XFS speculatively
+		// preallocates post-EOF blocks get reclaimed when volDev is
+		// closed.
+		// https://github.com/longhorn/longhorn/issues/2503
 		// We want to truncate regular files, but not device
 		if stat.Mode()&os.ModeType == 0 {
 			log.Debugf("Truncate %v to size %v", volDevName, vol.Size)
@@ -626,6 +631,12 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
 				return
 			}
 		}
+
+		if err := performIncrementalRestore(srcVolumeName, volDev, lastBackup, backup, bsDriver, config); err != nil {
+			deltaOps.UpdateRestoreStatus(volDevName, 0, err)
+			return
+		}
+
 		deltaOps.UpdateRestoreStatus(volDevName, PROGRESS_PERCENTAGE_BACKUP_TOTAL, nil)
 	}()
 	return nil


c3y1huang:
Test:
```
root@longhorn-test:/integration/tests# ./run.sh -x -s test_basic.py::test_dr_volume_with_backup_block_deletion test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress test_basic.py::test_dr_volume_with_all_backup_blocks_deleted
+ flake8
+ pytest -v -x -s test_basic.py::test_dr_volume_with_backup_block_deletion test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress test_basic.py::test_dr_volume_with_all_backup_blocks_deleted
================================== test session starts ===================================
platform linux -- Python 3.9.4, pytest-5.3.1, py-1.10.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /integration, inifile: pytest.ini
plugins: repeat-0.9.1
collected 6 items                                                                        

test_basic.py::test_dr_volume_with_backup_block_deletion[s3] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion[nfs] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress[s3] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress[nfs] PASSED
test_basic.py::test_dr_volume_with_all_backup_blocks_deleted[s3] PASSED
test_basic.py::test_dr_volume_with_all_backup_blocks_deleted[nfs] PASSED

============================= 6 passed in 669.08s (0:11:09) ==============================
```

deltablock.go
```diff
@@ -509,7 +509,7 @@ func restoreBlockToFile(volumeName string, volDev *os.File, bsDriver BackupStore
 	if _, err := volDev.Seek(blk.Offset, 0); err != nil {
 		return err
 	}
-	if _, err := io.CopyN(volDev, r, DEFAULT_BLOCK_SIZE); err != nil {
+	if _, err := volDev.Write(r); err != nil {
```
shuo-wu, line 512: Rename the variable `r`

deltablock.go
```diff
@@ -481,14 +489,6 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
...
 		}
 
-		// We want to truncate regular files, but not device
-		if stat.Mode()&os.ModeType == 0 {
-			log.Debugf("Truncate %v to size %v", volDevName, vol.Size)
-			if err := volDev.Truncate(vol.Size); err != nil {
-				deltaOps.UpdateRestoreStatus(volDevName, progress, err)
-				return
-			}
-		}
```
shuo-wu, line 491: One quick question, is it possible that this truncation after the writing is the culprit?

c3y1huang: Possibly not. I've verified the fix before moving the truncation.

innobead:
CopyN and Write implementation are different because copyn uses buffer, so we need to make sure no extra memory consumption caused by this change. (it seems fine in this case)

Also, why u changed the order of truncation operation? (It's better to explain here) thanks!

c3y1huang:
> Also, why u changed the order of truncation operation? (It's better to explain here) thanks!

Just for safety because I am not sure if it would cause side effects if the file doesn't already contain existing blocks.

deltablock.go
c3y1huang: Addressed.

PhanLe1010:
The code LGTM.
I haven't understood why this problem only happens with sles15-sp2. Do you have more insight @c3y1huang ?

yasker:
NOT LGTM.

I don't understand the reason why `io.CopyN` won't copy N bytes as specified in the parameter, and why it only exists in SLES.

This is a core logic in the backup store. I think the risk of changing it without understanding the root cause is high. Since it's not a regression, I don't consider this issue as a blocker to the release.

yasker (changes requested):
Need to understand the exact reason why this happens and only happens in SLES, before we can merge this.

innobead:
yes, this is a good catch actually, but we need more regression on that and undersand the root cause first.

c3y1huang:
Thanks all for the feedback. I will do more investigation.

c3y1huang:
The file block count is the same on SUSE Leap and RHEL/CentOS8 with a basic sample code to reproduce.

Running the e2e tests (before and after the fix), the behavior is the same on RHEL/CentOS 8.

c3y1huang:
SUSE Leap and SLES use XFS as the default filesystem, and CentOS 7 also. Before CentOS 7 was using ext4 and so is Ubuntu (20.04).

shuo-wu:
According to my test result, the DR volume data in `test_dr_volume_with_backup_block_deletion ` becomes inconsistent after [restoring backup2](https://github.com/longhorn/longhorn-tests/blob/526d88dff4fd50ebf8d690a19ba327d625d3fcd8/manager/integration/tests/test_basic.py#L784-L791). 

```
The written data length: 4194304 (4M)
The volume data length: 4194304 (4M)
The longest continuous 0 data in volume, length: 2031616 start: 2097152 (2M) end: 4128767

```

c3y1huang:
According to https://xfs.org/index.php/XFS_FAQ#Q:_Why_do_files_on_XFS_use_more_data_blocks_than_expected.3F
```
The XFS speculative preallocation algorithm allocates extra blocks beyond end of file (EOF) to minimize file fragmentation during buffered write workloads.
```
and
```
XFS speculatively preallocates post-EOF blocks on file extending writes in anticipation of future extending writes. The size of a preallocation is dynamic and depends on the runtime state of the file and fs.
In most cases, speculative preallocation is automatically reclaimed when a file is closed.
```
In our case, the volume file gets truncated before is closed and this moved the EOF causing unable to find and reclaim the speculative preallocation at post-EOF. And this resulted in the extra file block.

The issue manifested in the incremental restore because the first backup does not call for sparse.FoldFile and the fragmented blocks are not overwritten to the offset.

To ensure EOF is not moved to have post-EOF reclaimed by the file truncate, we should be able to resolve by adding truncate also before the file IO and keeping the io.CopyN as it is.

deltablock.go
shuo-wu: Finally, this one becomes the culprit...

c3y1huang: You are right all along, this one should be the culprit :smile: .

c3y1huang:
SLES15-SP2:
<details sles-test-dr-volume-backup>
<summary>test_dr_volume_*_backup</summary>
<p>

```
tbu
```

</p>
</details>  

<details sles-coretest>
<summary>coretest</summary>
<p>

```
tbu
```

</p>
</details> 

Ubuntu:
<details ubuntu-test-dr-volume-backup>
<summary>test_dr_volume_*_backup</summary>
<p>

tbu

</p>
</details>  

<details ubuntu-coretest>
<summary>coretest</summary>
<p>

tbu

</p>
</details>

shuo-wu (approved):

yasker:
Great work @c3y1huang ! I think we've found the root cause.

However, it a bit hard for me to understand the logic by reading https://github.com/longhorn/backupstore/pull/70#issuecomment-826649095 . Can we have a more detailed write-up regarding the explanation and solution, so anyone who reads it can understand what's happened? Also, make sure we have posted the explanation in the issue https://github.com/longhorn/longhorn/issues/2503 as well, so we can potentially refer it in the future.

Thanks.

yasker:
Also, what would be the potential impact of this issue? Is our support for XFS completely broken before this commit? If that's the case, we need to send messages to the communities and customers immediately regarding this, make sure they don't use XFS as the filesystem, or don't use backup/restore on XFS.

I know checksum is different, but not sure if it's harmful or not to the filesystem on top of it.

innobead (approved):

c3y1huang:
More detailed write-up in https://github.com/longhorn/longhorn/issues/2503#issuecomment-828158607.

deltablock.go
```diff
@@ -489,6 +502,7 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
 				return
 			}
 		}
+
```
yasker, line 505: Do we need to truncate the file again once the restore is done?

deltablock.go
```diff
@@ -626,6 +654,7 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
...
```
yasker, line 657: Same here.

yasker (changes requested):
Thanks for the detailed write-up @c3y1huang ! Now it's clear to me what's happened.

My only question is if it's necessary for us to truncate the file twice then.

deltablock.go
c3y1huang: The pre-truncate is to establish the volume EOF.

As for the post-truncate can be removed if the restore is definite to work as expected. I've kept this to ensure any file blocks exceeding EOF get removed even if the EOF got changed again. But I am not sure if there will be a case if data will be over the volume length. What do you think?

yasker: We should remove this truncate since we don't expect any write can be beyond the size of the file. Otherwise, it's a bug.

If by removing this unnecessary `truncate` we found other bugs, that would be good too.

c3y1huang:
SLES15-SP2:
<details sles-test-dr-volume-backup>
<summary>test_dr_volume_*_backup</summary>
<p>

```
ip-172-30-0-211:/home/ec2-user # kubectl exec -it longhorn-test -- bash
root@longhorn-test:/integration/tests# ./run.sh -x -s test_basic.py::test_dr_volume_with_backup_block_deletion test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress test_basic.py::test_dr_volume_with_all_backup_blocks_deleted
+ flake8
+ pytest -v -x -s test_basic.py::test_dr_volume_with_backup_block_deletion test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress test_basic.py::test_dr_volume_with_all_backup_blocks_deleted
============================= test session starts ==============================
platform linux -- Python 3.9.4, pytest-5.3.1, py-1.10.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /integration, inifile: pytest.ini
plugins: repeat-0.9.1
collecting ... collected 6 items

test_basic.py::test_dr_volume_with_backup_block_deletion[s3] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion[nfs] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress[s3] PASSED
test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress[nfs] PASSED
test_basic.py::test_dr_volume_with_all_backup_blocks_deleted[s3] PASSED
test_basic.py::test_dr_volume_with_all_backup_blocks_deleted[nfs] PASSED

======================== 6 passed in 683.89s (0:11:23) =========================
```

</p>
</details>  

<details sles-coretest>
<summary>coretest</summary>
<p>

```
root@longhorn-test:/integration/tests# ./run.sh -x -s -m coretest
+ flake8

+ pytest -v -x -s -m coretest
============================= test session starts ==============================
platform linux -- Python 3.9.4, pytest-5.3.1, py-1.10.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /integration, inifile: pytest.ini
plugins: repeat-0.9.1
collecting ... collected 232 items / 177 deselected / 55 selected

test_backing_image.py::test_backing_image_basic_operation PASSED
test_backing_image.py::test_backing_image_content PASSED
test_backing_image.py::test_volume_basic_with_backing_image PASSED
test_backing_image.py::test_volume_iscsi_basic_with_backing_image PASSED
test_backing_image.py::test_snapshot_with_backing_image PASSED
test_backing_image.py::test_backup_with_backing_image PASSED
test_backing_image.py::test_backup_labels_with_backing_image PASSED
test_backing_image.py::test_ha_simple_recovery_with_backing_image PASSED
test_backing_image.py::test_engine_live_upgrade_with_backing_image PASSED
test_backing_image.py::test_csi_io_with_backing_image PASSED
test_backing_image.py::test_csi_backup_with_backing_image PASSED
test_basic.py::test_hosts PASSED
test_basic.py::test_settings PASSED
test_basic.py::test_volume_basic PASSED
test_basic.py::test_snapshot PASSED
test_basic.py::test_backup PASSED
test_basic.py::test_backup_labels PASSED
test_basic.py::test_restore_inc[s3] PASSED
test_basic.py::test_restore_inc[nfs] PASSED
test_basic.py::test_listing_backup_volume PASSED
test_basic.py::test_volume_multinode PASSED
test_basic.py::test_volume_scheduling_failure PASSED
test_basic.py::test_setting_default_replica_count PASSED
test_basic.py::test_volume_update_replica_count PASSED
test_basic.py::test_attach_without_frontend PASSED
test_basic.py::test_storage_class_from_backup[s3] PASSED
test_basic.py::test_storage_class_from_backup[nfs] PASSED
test_basic.py::test_expansion_basic PASSED
test_basic.py::test_restore_inc_with_expansion[s3] PASSED
test_basic.py::test_restore_inc_with_expansion[nfs] PASSED
test_basic.py::test_expansion_canceling PASSED
test_basic.py::test_running_volume_with_scheduling_failure PASSED
test_basic.py::test_expansion_with_scheduling_failure PASSED
test_basic.py::test_allow_volume_creation_with_degraded_availability PASSED
test_basic.py::test_allow_volume_creation_with_degraded_availability_error PASSED
test_csi.py::test_csi_mount PASSED
test_csi.py::test_csi_offline_expansion PASSED
test_csi.py::test_csi_expansion_with_replica_failure PASSED
test_csi.py::test_allow_volume_creation_with_degraded_availability_csi PASSED
test_engine_upgrade.py::test_engine_offline_upgrade PASSED
test_engine_upgrade.py::test_engine_live_upgrade PASSED
test_engine_upgrade.py::test_engine_live_upgrade_with_intensive_data_writing PASSED
test_ha.py::test_ha_simple_recovery PASSED
test_ha.py::test_ha_salvage PASSED
test_ha.py::test_single_replica_failed_during_engine_start PASSED
test_migration.py::test_migration_confirm PASSED
test_migration.py::test_migration_rollback PASSED
test_node.py::test_update_node PASSED
test_node.py::test_node_disk_update PASSED
test_node.py::test_replica_scheduler_no_disks PASSED
test_node.py::test_node_controller_sync_storage_available PASSED
test_node.py::test_node_controller_sync_disk_state PASSED
test_node.py::test_replica_datapath_cleanup PASSED
test_provisioner.py::test_provisioner_mount PASSED
test_statefulset.py::test_statefulset_scaling PASSED

========== 55 passed, 177 deselected, 1 warning in 6507.11s (1:48:27) ==========
```

</p>
</details> 

Ubuntu:
<details ubuntu-test-dr-volume-backup>
<summary>test_dr_volume_*_backup</summary>
<p>

tbu

</p>
</details>  

<details ubuntu-coretest>
<summary>coretest</summary>
<p>

tbu

</p>
</details>

deltablock.go
c3y1huang: Got it. Addressed.

yasker (approved):
LGTM.

shuo-wu (approved):

innobead (approved):

PR #73, opened 2021-06-17
Fix list in progress backup panic

jenting:
#### Proposal Change
Returns error at in-progress backup, so we could check do ignore it _or_ include it when all the jobs finished.
Add unit test to cover the in-progress backup and faulty backup cases.

#### Linked Issue
https://github.com/longhorn/longhorn/issues/2683

Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>

### DIFF
--- a/list.go
+++ b/list.go
@@ -108,7 +108,7 @@ func addListVolume(volumeName string, driver BackupStoreDriver, volumeOnly bool)
 				info = failedBackupInfo(backupName, volumeName, driver.GetURL(), err)
 			} else if isBackupInProgress(backup) {
 				// for now we don't return in progress backups to the ui
-				return nil, fmt.Errorf("Backup in progress")
+				return nil, fmt.Errorf("backup %v is still in progress", backup.Name)
 			} else {
 				info = fillBackupInfo(backup, driver.GetURL())
 			}
--- a/list.go
+++ b/list.go
@@ -108,6 +108,7 @@ func addListVolume(volumeName string, driver BackupStoreDriver, volumeOnly bool)
 				info = failedBackupInfo(backupName, volumeName, driver.GetURL(), err)
 			} else if isBackupInProgress(backup) {
 				// for now we don't return in progress backups to the ui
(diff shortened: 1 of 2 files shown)


list_test.go
```diff
@@ -228,7 +243,7 @@ func TestListSingleVolumeBackups(t *testing.T) {
 	for volumeName := range m.volumeInfos {
 		volumeInfo, err := List(volumeName, mockDriverURL, false)
 		assert.NoError(err)
-		assert.Equal(m.numOfBackups, len(volumeInfo[volumeName].Backups))
+		assert.Equal(m.numOfBackups-m.numOfInProgressBackups, len(volumeInfo[volumeName].Backups))
```
shuo-wu, line 246: Should we check faulty backups as well?

list.go
```diff
@@ -108,6 +108,7 @@ func addListVolume(volumeName string, driver BackupStoreDriver, volumeOnly bool)
 				info = failedBackupInfo(backupName, volumeName, driver.GetURL(), err)
 			} else if isBackupInProgress(backup) {
 				// for now we don't return in progress backups to the ui
+				return nil, fmt.Errorf("Backup in progress")
```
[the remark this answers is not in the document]
jenting: I'd want to make it the same as https://github.com/longhorn/backupstore/blob/89bd8d2/list.go#L255

list_test.go
jenting: Good suggestion!

list.go
```diff
@@ -118,7 +119,10 @@ func addListVolume(volumeName string, driver BackupStoreDriver, volumeOnly bool)
 	}
 
 	for _, tracker := range trackers {
-		payload, _ := tracker.Result()
+		payload, err := tracker.Result()
+		if err != nil {
+			continue
```
innobead, line 124: Can we at least dump the error log here? debug or trace.

innobead (approved):
LGTM

PR #75, opened 2021-07-12
Add new API to get the config last modification time

jenting:
#### Propose Changes

- Rename backup volume config and backup config configuration-related code to use the word `config` instead of `metadata` to make it more clear.
- `backup config-metadata`: a new command to get the config (backup volume config _or_ backup config) metadata to get the last modification time. This could prevent the longhorn manager read the config periodically even the config did not change.

#### Issue

https://github.com/longhorn/longhorn/issues/1761

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -96,7 +96,7 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
 	return nil
 }
 
-func EncodeMetadataURL(backupName, volumeName, destURL string) string {
+func EncodeConfigURL(backupName, volumeName, destURL string) string {
 	v := url.Values{}
 	v.Add("volume", volumeName)
 	if backupName != "" {
@@ -105,7 +105,7 @@ func EncodeMetadataURL(backupName, volumeName, destURL string) string {
 	return destURL + "?" + v.Encode()
 }
 
-func DecodeMetadataURL(backupURL string) (string, string, string, error) {
+func DecodeConfigURL(backupURL string) (string, string, string, error) {
 	u, err := url.Parse(backupURL)
 	if err != nil {
 		return "", "", "", err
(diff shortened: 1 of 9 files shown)


shuo-wu (commented):
The naming is weird... The config file store the meta info of a backup or a backup volume. But here `metadata` means the meta info of the config file itself and it reports the last modified time only...

jenting:
> The naming is weird... The config file store the meta info of a backup or a backup volume. But here `metadata` means the meta info of the config file itself and it reports the last modified time only...

If using the `head` naming, would it more intuitive? https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

cmd/head.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/urfave/cli"
+
+	"github.com/longhorn/backupstore"
+	"github.com/longhorn/backupstore/util"
+)
+
+func GetConfigMetadataCmd() cli.Command {
+	return cli.Command{
+		Name:   "head",
+		Usage:  "get the config metadata",
```
shuo-wu, line 15: `... this returns the last modification time of a config file for now.`

backupstore.go
```diff
@@ -96,7 +96,7 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
 	return nil
 }
 
-func EncodeMetadataURL(backupName, volumeName, destURL string) string {
+func EncodeConfigURL(backupName, volumeName, destURL string) string {
```
shuo-wu, line 99: `EncodeBackupURL`. Similar below

shuo-wu (changes requested):

cmd/head.go
jenting: Add to the `description`.

shuo-wu (approved):
LGTM. Remember to rearrange the commits

PR #78, opened 2021-07-26
bump go version to 1.16 and update dependencies.

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/2838

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>


Dockerfile.dapper
```diff
@@ -1,25 +1,29 @@
...
-    DOCKER_URL_arm=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm \
+# Docker
+ENV DOCKER_URL_amd64=https://download.docker.com/linux/ubuntu/dists/focal/pool/stable/amd64/docker-ce-cli_20.10.7~3-0~ubuntu-focal_amd64.deb \
+    DOCKER_URL_arm64=https://download.docker.com/linux/ubuntu/dists/focal/pool/stable/arm64/docker-ce-cli_20.10.7~3-0~ubuntu-focal_arm64.deb \
     DOCKER_URL=DOCKER_URL_${ARCH}
 
-RUN wget -O - ${!DOCKER_URL} > /usr/bin/docker && chmod +x /usr/bin/docker
+RUN wget ${!DOCKER_URL} -O docker_ce_${ARCH} && dpkg -i docker_ce_${ARCH}
 
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
```
jenting, line 22: Not quite understand this env `GOLANG_ARCH_arm=armv6l`.

Dockerfile.dapper
```diff
@@ -1,25 +1,29 @@
...
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.13.3.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
-    go get github.com/rancher/trash && go get -u golang.org/x/lint/golint
```
shuo-wu, line 22: Is `trash` no longer required?

shuo-wu (changes requested):
Please rename the commit or separate the changes into multiple commits.

joshimoo:
Just to be aware go 1.16 has some module handling changes. So if you update your local go version a go get will look quite a bit different :)

https://blog.golang.org/go116-module-changes

Generally updating is a good idea once we upgrade the kubernetes version https://github.com/longhorn/longhorn/issues/2757
@meldafrawi please also create a Task issue in the longhorn/longhorn repo and reference it here as part of the PR.

Dockerfile.dapper
joshimoo: I think this is just copy pasta from the rancher side, we don't build `armv6` images but the code can still be compiled on `armv6`

Dockerfile.dapper
meldafrawi: IMO, no.
According to trash docs, it's used to manage `vendor` directory, which is already achieved using `go mod` commands.

Dockerfile.dapper
meldafrawi: @jenting, I've changed arm architecture  to `arm64`.

Dockerfile.dapper
shuo-wu: But it seems that `trash` is still used in Makefile. Can we remove `trash` from Makefile as well?

PR #81, opened 2021-09-08
Set NFS mount option `actimeo` to 1 second

jenting:
#### Proposal Change

The NFS server has the cache.
If we have more than one node mount the same NFS server, the cache coherence problem occurs among these NFS client nodes.
    
Therefore, we set mount option `actimeo` to 1 second, it sets all min/max time of NFS client cache attributes of a file/directory before it requests fresh attribute information from a server.
    
Ref to:
- https://linux.die.net/man/5/nfs
- https://serverfault.com/a/640600

#### Issue

https://github.com/longhorn/longhorn/issues/2926

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -83,7 +83,7 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 func (b *BackupStoreDriver) mount() (err error) {
 	defer func() {
 		if err != nil {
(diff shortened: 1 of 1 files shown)


joshimoo (approved):
LGTM, setting actimeo to do attribute lookup seems like a good idea.

Have you considered using `lookupcache=none` to disable directory entry caching.
This way every creation/deletion of a file will be present immediately.

Since each lookup will be handled by the server.

PR #85, opened 2021-10-29
nfs: only return coalesce error when last NFSv4 version fails to mount

derekbit:
(1) Only return coalesce error when last NFSv4 version fails to mount.
(2) Remove NFSv3 try mount and umount from defer function.

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -81,15 +81,11 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 }
 
 func (b *BackupStoreDriver) mount() (err error) {
+	var errs []string
+
 	defer func() {
 		if err != nil {
-			if _, err := util.Execute("mount", []string{"-t", "nfs", "-o", "nfsvers=3", "-o", "nolock", "-o", "actimeo=1", b.serverPath, b.mountDir}); err != nil {
-				return
-			}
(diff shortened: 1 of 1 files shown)


nfs/nfs.go
```diff
@@ -81,27 +81,26 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
 	}()
 
 	if !util.IsMounted(b.mountDir) {
 		for _, version := range MinorVersions {
 			log.Debugf("attempting mount for nfs path %v with nfsvers %v", b.serverPath, version)
 			_, err = util.Execute("mount", []string{"-t", "nfs4", "-o", fmt.Sprintf("nfsvers=%v", version), "-o", "actimeo=1", b.serverPath, b.mountDir})
-			if err == nil || !strings.Contains(err.Error(), UnsupportedProtocolError) {
+			if err == nil {
 				break
+			} else {
```
jenting, line 98: The else statement is redundant.

nfs/nfs.go
```diff
@@ -81,27 +81,26 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
+				errs = append(errs, fmt.Sprintf("ver=%v: %v", version, err.Error()))
 			}
 		}
 	}
+
```
jenting, line 103: Need to check if len(errs) > 0, then transfer error array to error and return it.

jenting (changes requested):

PR #87, opened 2021-10-30
nfs: only return coalesce error when last NFSv4 version fails to mount

derekbit:
(1) Only return coalesce error when last NFSv4 version fails to mount.
(2) Remove NFSv3 try mount/umount from defer function.

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -81,36 +81,23 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 }
 
 func (b *BackupStoreDriver) mount() (err error) {
-	defer func() {
-		if err != nil {
-			if _, err := util.Execute("mount", []string{"-t", "nfs", "-o", "nfsvers=3", "-o", "nolock", "-o", "actimeo=1", b.serverPath, b.mountDir}); err != nil {
-				return
-			}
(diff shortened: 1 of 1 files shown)


nfs/nfs.go
```diff
@@ -81,36 +81,25 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
-	return err
-}
 
-func (b *BackupStoreDriver) unmount() error {
-	var err error
-	if util.IsMounted(b.mountDir) {
-		_, err = util.Execute("umount", []string{b.mountDir})
+	if len(errs) > 0 {
+		return errors.New(fmt.Sprintf("Cannot mount using NFSv4: %s", strings.Join(errs, ";")))
 	}
```
jenting, line 100:
```suggestion
	if util.IsMounted(b.mountDir) {
	    return nil
	}
	
	for _, version := range MinorVersions {
		log.Debugf("attempting mount for nfs path %v with nfsvers %v", b.serverPath, version)
		_, err = util.Execute("mount", []string{"-t", "nfs4", "-o", fmt.Sprintf("nfsvers=%v", version), "-o", "actimeo=1", b.serverPath, b.mountDir})
		if err == nil {
			return nil
		}
		err = errors.Wrapf(err, "ver=%s", version)
	}
        return err
```

jenting (approved):

innobead (approved):
LGTM

PR #89, opened 2022-09-06
feat(nfs): test connection

FrankYang0529:
If we lose NFS connection after setting up backup-target and then we try to initialize NFS BackupStoreDriver again in another program, the program may be stuck in [`MkdirAll`](https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/nfs/nfs.go#L67) or [`List`](https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/nfs/nfs.go#L74). So I use `util.Execute` to avoid being stuck in `mkdir` command.

Issue: https://github.com/harvester/harvester/issues/2631

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -3,9 +3,9 @@ package nfs
 import (
 	"fmt"
 	"net/url"
-	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/longhorn/backupstore"
 	"github.com/longhorn/backupstore/fsops"
@@ -64,7 +64,7 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 
 	b.serverPath = u.Host + u.Path
 	b.mountDir = filepath.Join(MountDir, strings.TrimRight(strings.Replace(u.Host, ".", "_", -1), ":"), u.Path)
-	if err := os.MkdirAll(b.mountDir, os.ModeDir|0700); err != nil {
+	if _, err = util.ExecuteWithCustomTimeout("mkdir", []string{"-m", "700", "-p", b.mountDir}, 3*time.Second); err != nil {
 		return nil, fmt.Errorf("Cannot create mount directory %v for NFS server: %v", b.mountDir, err)
 	}
 
--- a/util/util.go
+++ b/util/util.go
@@ -127,11 +127,20 @@ func ValidateName(name string) bool {
 }
 
 func Execute(binary string, args []string) (string, error) {
-	var output []byte
-	var err error
-
 	ctx, cancel := context.WithTimeout(context.Background(), cmdTimeout)
 	defer cancel()
+	return execute(ctx, binary, args)
+}
+
+func ExecuteWithCustomTimeout(binary string, args []string, timeout time.Duration) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	return execute(ctx, binary, args)
+}
+
+func execute(ctx context.Context, binary string, args []string) (string, error) {
+	var output []byte
+	var err error
 
 	cmd := exec.CommandContext(ctx, binary, args...)
 	done := make(chan struct{})


bk201:
@shuo-wu The PR tries to let the call fail fast if NFS server is offline. If NFS server is off, The first `os.MkdirAll` call could stuck because it stats mount point first.

nfs/nfs.go
```diff
@@ -111,3 +117,12 @@ func (b *BackupStoreDriver) GetURL() string {
...
 }
+
+func testConnection(url string) error {
+	conn, err := net.Dial("tcp", url)
+	if err != nil {
+		return err
+	}
+	conn.Close()
+	return nil
+}
```
derekbit, line 128: Is it better to add a timeout?

nfs/nfs.go
```diff
@@ -62,6 +63,11 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 		return nil, fmt.Errorf("Cannot find nfs path")
 	}
 
+	url := u.Host + "111"
```
derekbit, line 66: Just curious. What does "111" mean?

Vicente-Cheng: port 111 is for `portmapper`
maybe we could check port 111 and port 2049 (2049 is for `nfs`)

derekbit: Got it. Thank you.
Suggest defining a const for the port number.

Vicente-Cheng:
For the port checking, maybe we could try to use `rpcinfo` command?

Another question is that even if the listening ports are what we want, we may still hang on the `MkdirAll` or `List`.

Should we need to another checking for target mountpoint? (Like check export by `showmount` or do some operations like `stat` on target mountpoint with timeout)

FrankYang0529:
Hi @Vicente-Cheng and @derekbit, thanks for your reviews and suggestions. I am thinking how about using `util.Execute`, so we don't need to check connection and we can avoid hanging on the `MkdirAll`. WDYT? Thank you.

derekbit:
> Hi @Vicente-Cheng and @derekbit, thanks for your reviews and suggestions. I am thinking about using `util.Execute` for `Mkdirall`, so we don't need to check connection and we can avoid hanging on the `MkdirAll`. WDYT? Thank you.

Sounds good. :)

Vicente-Cheng:
> Hi @Vicente-Cheng and @derekbit, thanks for your reviews and suggestions. I am thinking about using [`util.Execute`](https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/util/util.go#L129-L156) for `Mkdirall`, so we don't need to check connection and we can avoid hanging on the `MkdirAll`. WDYT? Thank you.

Sure, sounds good with a timeout!

nfs/nfs.go
FrankYang0529: Removed `testConnection`. Thank you.

nfs/nfs.go
FrankYang0529: Removed `testConnection`. Thank you.

nfs/nfs.go
```diff
@@ -64,7 +63,7 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 
 	b.serverPath = u.Host + u.Path
 	b.mountDir = filepath.Join(MountDir, strings.TrimRight(strings.Replace(u.Host, ".", "_", -1), ":"), u.Path)
-	if err := os.MkdirAll(b.mountDir, os.ModeDir|0700); err != nil {
+	if _, err = util.Execute("mkdir", []string{"-m", "700", "-p", b.mountDir}); err != nil {
 		return nil, fmt.Errorf("Cannot create mount directory %v for NFS server: %v", b.mountDir, err)
```
derekbit, line 67: Just thought of one case that if the NFS folder is somehow unmounted, then the mkdir still works.
Probably can check the mount table as well for recomfirming the NFS mountpoint is still alive.

FrankYang0529: Yeah, I think we will do it here.

https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/nfs/nfs.go#L83-L101

util/util.go
```diff
@@ -23,7 +23,7 @@ const (
 )
 
 var (
-	cmdTimeout = time.Minute // one minute by default
+	cmdTimeout = 15 * time.Second // 15 seconds by default
```
bk201, line 26: Suggest not changing the timeout; many code paths use this value. (e.g., upload).

FrankYang0529: Oh! Yeah! You are right. Thanks for catching this. How about adding a new `ExcuteWithCustomTimeout` function?

FrankYang0529: Updated it. Thank you.

innobead (approved):

derekbit:
> @derekbit Could you help update the dependency of manager, instance manager, and engine? Thanks.

Sure thing!

PR #91, opened 2022-10-20
fix(nfs): set mount timeout

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/2986

### DIFF
--- a/nfs/nfs.go
+++ b/nfs/nfs.go
@@ -15,8 +15,9 @@ import (
 )
 
 var (
-	log           = logrus.WithFields(logrus.Fields{"pkg": "nfs"})
-	MinorVersions = []string{"4.2", "4.1", "4.0"}
+	log            = logrus.WithFields(logrus.Fields{"pkg": "nfs"})
+	MinorVersions  = []string{"4.2", "4.1", "4.0"}
+	defaultTimeout = 5 * time.Second
 )
 
 type BackupStoreDriver struct {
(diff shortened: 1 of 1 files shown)


innobead (commented):
LGTM, but I would like to see @derekbit @shuo-wu review this.

nfs/nfs.go
```diff
@@ -89,7 +89,7 @@ func (b *BackupStoreDriver) mount() (err error) {
 
 	for _, version := range MinorVersions {
 		log.Debugf("attempting mount for nfs path %v with nfsvers %v", b.serverPath, version)
-		_, err = util.Execute("mount", []string{"-t", "nfs4", "-o", fmt.Sprintf("nfsvers=%v", version), "-o", "actimeo=1", b.serverPath, b.mountDir})
+		_, err = util.ExecuteWithCustomTimeout("mount", []string{"-t", "nfs4", "-o", fmt.Sprintf("nfsvers=%v", version), "-o", "actimeo=1", b.serverPath, b.mountDir}, 3*time.Second)
```
derekbit, line 92: Suggest making `3*time.Second` as a constant.

shuo-wu (commented):
LGTM except for Derek's comment.

nfs/nfs.go
FrankYang0529: Updated it. Thank you.

shuo-wu (approved):

derekbit (approved):

innobead:
@derekbit could you help update this dependency to other depender repos? master and 1.3 only.

PR #92, opened 2022-12-06
refactor(system-backup/restore): define SystemBackups

c3y1huang:
https://github.com/longhorn/longhorn-manager/pull/1489#discussion_r1037195104
https://github.com/longhorn/longhorn/issues/1455


innobead (approved):
LGTM

PR #94, opened 2023-01-03
Improve backup performance with faster compression algorithm and multiple threads

derekbit:
Longhorn/longhorn#5189


deltablock.go
```diff
@@ -885,62 +899,182 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
...
+	go func() {
+		defer close(errChan)
+
+		volDevName := config.Filename
+		volDev, err := os.OpenFile(volDevName, os.O_RDWR, 0666)
+		if err != nil {
+			errChan <- err
+			return
+		}
+		defer volDev.Close()
```
shuo-wu, line 1025: Does this mean that the file will be opened and operated by multiple threads simultaneously? Will this be a potential issue? Should we open the file in the caller side instead?

deltablock.go
```diff
@@ -885,62 +899,182 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
...
-			if err := fillBlockToFile(&emptyBlock, volDev, lB.Offset); err != nil {
-				return err
+		}
+	}()
+
+	return blockChan, errChan
+}
+
+func populateBlocksForDeltaBlockBackup(bsDriver BackupStoreDriver, backup *Backup) (<-chan *Block, <-chan error) {
+	blockChan := make(chan *Block, 1)
```
shuo-wu, line 969: NIT: Can increasing the channel buffer size help improve the performance? Or are there any other concerns? Similar logic applies to the backup part.

deltablock.go
```diff
@@ -885,62 +899,182 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
...
 
-func performIncrementalRestore(bsDriver BackupStoreDriver, srcVolumeName string, volDev *os.File, lastBackup *Backup, backup *Backup, config *DeltaRestoreConfig) error {
-	var progress int
-	volDevName := config.Filename
-	deltaOps := config.DeltaOps
+type Block struct {
+	offset            int64
+	blockChecksum     string
+	compressionMethod string
+	zeros             bool
```
shuo-wu, line 906: NIT: `isZeroBlock`

deltablock.go
```diff
@@ -885,62 +899,182 @@ func RestoreDeltaBlockBackupIncrementally(config *DeltaRestoreConfig) error {
...
-		} else {
-			if err := fillBlockToFile(&emptyBlock, volDev, lB.Offset); err != nil {
-				return err
+		}
+	}()
+
+	return blockChan, errChan
+}
+
+func populateBlocksForDeltaBlockBackup(bsDriver BackupStoreDriver, backup *Backup) (<-chan *Block, <-chan error) {
```
shuo-wu, line 968: `populateBlocksForFullRestore`

deltablock.go
```diff
@@ -276,6 +279,47 @@ func getProgress(total, processed int64) int {
...
+
+func updateStatis(deltaBackup *Backup, progress *backupProgress, checksum string, newBlock bool) {
+	processingBlocks := deltaBackup.ProcessingBlocks
+
+	processingBlocks.Lock()
+	defer processingBlocks.Unlock()
+
+	// Update deltaBackup.Blocks
+	blocks := processingBlocks.blocks[checksum]
+	for _, block := range blocks {
```
shuo-wu, line 311: 1. It seems that there can be multiple blocks holding the same checksum when the blocks of mapping have the same content, right?
Then the next question is, why not using offset+length as the key? Then `processingBlocks.blocks` can be `map[string]*BlockMapping{}`
2. Actually, this function is invoked by `processBlock`, which means single block handling. Does we need to consider the multi-block case here?

shuo-wu (commented):
To be honest, the naming and layering of the backup part are a little bit hard-to-understand. Not sure if we can improve it...

derekbit:
> To be honest, the naming and layering of the backup part are a little bit hard-to-understand. Not sure if we can improve it...

Do you mean the whole backup layer?

deltablock.go
derekbit: Increasing the channel size is fine.

deltablock.go
derekbit: Each goroutine opens create a file descriptor for the same disk file. Each file descriptor's operations are independent and won't be an issue.

deltablock.go
```diff
@@ -285,21 +329,18 @@ func processBlock(bsDriver BackupStoreDriver, config *DeltaBackupConfig,
 	deltaOps := config.DeltaOps
 
 	checksum := util.GetChecksum(block)
+
+	if isBlockBeingProcessing(deltaBackup, offset, checksum) {
+		return nil
+	}
```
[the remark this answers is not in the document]
derekbit: Not catch you question.
One block is handled by one goroutine.

deltablock.go
```diff
@@ -238,77 +250,237 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+	go func() {
+		defer close(mappingChan)
+		defer close(errChan)
+
+		for _, mapping := range delta.Mappings {
+			if mapping.Size%delta.BlockSize != 0 {
+				errChan <- fmt.Errorf("mapping size %v is not multiples of backup block size %v",
+					mapping.Size, delta.BlockSize)
+				return
+			}
```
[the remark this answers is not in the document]
derekbit: You are right. We can remove it.

deltablock.go
derekbit: Updated. Thank you.

deltablock.go
derekbit: Updated. Thank you.

deltablock.go
derekbit: Updated it to 10. Thank you.

deltablock.go
shuo-wu: OK, finally I see the idea of this part... You are trying to avoid uploading blocks with the same checksum repeatedly... I was confused by the multi-layer invocation (mapping-block-processing block-....) in the 1st round...
But the next question is, after this function deletes the checksum record from `processingBlocks.blocks`, we will re-do the uploading right? Does this mean that this processing check mechanism can avoid duplicate uploading **in a pretty short period**?

shuo-wu:
I mean all functions including the functions you already mentioned and the new functions you introduced...

derekbit:
> I mean all functions including the functions you already mentioned and the new functions you introduced...

For the new functions
```
backup: backupMappings->backupMapping->backupBlock->isBlockBeingProcessed->updateBlocksAndProgress
restore: restoreBlocks->restoreBlock->fillZeros or restoreBlockToFile
```
For the exsting functions, for example, we can merge `RestoreDeltaBlockBackupIncrementally` and `RestoreDeltaBlockBackup` in another refactoring ticket.

deltablock.go
derekbit: No. After adding a block to `processingBlocks.blocks`, if the block is not in `processingBlocks.blocks`, the next step is to look up (`FileExists()`) the block in the remote backupstore. If YES, the re-uploading should not happen.

systembackup/system_backup.go
```diff
@@ -78,14 +79,14 @@ func Download(localFilePath string, cfg *Config) error {
...
-	err = backupstore.SaveBackupStoreToLocalFile(remoteBackupURI, localFilePath, driver)
+	err = backupstore.SaveBackupStoreToLocalFile(driver, remoteBackupURI, localFilePath)
 	if err != nil {
 		return err
 	}
 
 	sha512sum, err := util.GetFileChecksum(localFilePath)
 	if err != nil {
-		return fmt.Errorf("failed to get %v checksum: %w", localFilePath, err)
+		return errors.Wrapf(err, "failed to get %v checksum", localFilePath)
```
joshimoo, line 89: Change is fine, I think we mostly use errors.Wrapf but for 
Knowledge sharing: the %w param will add the wrap/unwrap for the errors same as errors.Wrapf

deltablock.go
```diff
@@ -88,7 +94,8 @@ type DeltaRestoreOperations interface {
 }
 
 const (
-	DEFAULT_BLOCK_SIZE = 2097152
+	DEFAULT_BLOCK_SIZE        = 2097152
```
joshimoo, line 97: NIT: I find 2 * 1024 * 1024 easier to read, and immediately obvious that we are talking about 2MB

deltablock.go
```diff
@@ -122,29 +129,35 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
 		return "", false, err
 	}
 
-	defer lock.Unlock()
 	if err := lock.Lock(); err != nil {
 		return "", false, err
 	}
+	defer lock.Unlock()
```
joshimoo, line 135: NIT: is it  save to move this?

joshimoo (commented):
I like the refactoring and error message clean up, let's move them into a prerequisite separate PR that we can wave through quickly so we can keep the PR and review focused on the actual backup compression changes.

deltablock.go
derekbit: Ah, tested it before. Forgot moving it back.
But why the derfer function is put before `lock.Lock()`.

deltablock.go
derekbit: Good idea.

derekbit: Updated. Thank you.

derekbit:
> I like the refactoring and error message clean up, let's move them into a prerequisite separate PR that we can wave through quickly so we can keep the PR and review focused on the actual backup compression changes.

Move the refactoring codes to https://github.com/longhorn/backupstore/pull/95. Thank you.

deltablock.go
shuo-wu: I see. Then please add comments for line 331. For example:
```
This prevents multiple goroutines from trying to upload blocks that contain identical contents (with the same checksum but different offsets). 
After uploading, `bsDriver.FileExists(blkFile)` is used to avoid repeat uploading.
```

joshimoo:
@derekbit merged the refactoring you can `rebase -i` to drop the duplicate refactoring commits. Thanks for moving them out into it's own PR :)

deltablock.go
derekbit: Added. Thank you.

shuo-wu (approved):
LGTM.
Maybe we can re-organize `deltablock.go` in the future. For now let's focus on the feature itself.

innobead (commented):
In general, LGTM. Just a few questions/comments.

innobead:
> LGTM. Maybe we can re-organize `deltablock.go` in the future. For now let's focus on the feature itself.

ref: https://github.com/longhorn/longhorn/issues/5247 If anything we would like to improve, just create an issue for that.

deltablock.go
```diff
@@ -486,20 +727,27 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
...
-		blkCounts := len(backup.Blocks)
-		for i, block := range backup.Blocks {
-			log.Debugf("Restore for %v: block %v, %v/%v", volDevName, block.BlockChecksum, i+1, blkCounts)
-			if err := restoreBlockToFile(srcVolumeName, volDev, bsDriver, block); err != nil {
-				deltaOps.UpdateRestoreStatus(volDevName, progress, err)
-				return
-			}
-			progress = int((float64(i+1) / float64(blkCounts)) * PROGRESS_PERCENTAGE_BACKUP_SNAPSHOT)
-			deltaOps.UpdateRestoreStatus(volDevName, progress, err)
+		ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
derekbit: We don't have cancellation mechanism for backup and restore. For the stuck issue, we can have add new APIs for letting users stop them.

deltablock.go
```diff
@@ -239,77 +255,296 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+	if err := saveBackup(bsDriver, &Backup{
+		Name:              deltaBackup.Name,
+		VolumeName:        deltaBackup.VolumeName,
+		CompressionMethod: volume.CompressionMethod,
+		CreatedTime:       "",
+	}); err != nil {
+		return 0, "", err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
derekbit: Same here https://github.com/longhorn/backupstore/pull/94#discussion_r1065754247.

deltablock.go
```diff
@@ -239,77 +255,296 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+		}
+		totalBlockCounts += d.Size / delta.BlockSize
+	}
+	return totalBlockCounts, nil
+}
+
+// mergeErrorChannels will merge all error channels into a single error out channel.
+// the error out channel will be closed once the ctx is done or all error channels are closed
+// if there is an error on one of the incoming channels the error will be relayed.
+func mergeErrorChannels(ctx context.Context, channels ...<-chan error) <-chan error {
```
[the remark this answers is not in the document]
derekbit: Will address in https://github.com/longhorn/backupstore/pull/94#issuecomment-1377193085

test/backup_test.go
```diff
@@ -279,329 +279,347 @@ func (s *TestSuite) waitForRestoreCompletion(c *C, deltaOps *RawFileVolume) {
...
-
-	for i := int64(0); i < blockSize; i++ {
-		data[i] = letterBytes[rand.Intn(len(letterBytes))]
-	}
-	for i := int64(1); i < volumeContentSize/blockSize; i++ {
-		for j := int64(0); j < blockSize; j++ {
-			data[i*blockSize+j] = data[j]
-		}
-	}
+	compressionMethods := []string{"none"}
```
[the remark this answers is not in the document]
derekbit: Added. Thank you.

deltablock.go
```diff
@@ -239,77 +255,296 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (string, bool, error) {
...
+	}
+
+	// Update progress
+	progress.Lock()
+	if newBlock {
+		progress.newBlockCounts++
+	}
+	progress.processedBlockCounts += int64(len(blocks))
+	progress.progress = getProgress(progress.totalBlockCounts, progress.processedBlockCounts)
+	progress.Unlock()
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

deltablock.go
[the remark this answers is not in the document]
innobead: I see. The backup can be deleted by the backup command but how about restoring the stuck volume? I am curious if those multiple goroutines for backup/restore would be pending for there. 

However, this should be a different topic to tackle, because the original design is the same but just a single thread.

innobead:
BTW, lz4 lib is interesting. It uses assembly code for each different arch to compile together. 
https://github.com/pierrec/lz4/blob/v4/internal/lz4block/decode_amd64.s

However, it seems not to support the s390x arch. Will this be a concern? or can we just fallback to gzip for unsupported arches?

deltablock.go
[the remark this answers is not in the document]
derekbit: For restore stuck, it seems deleting the volume being restored is a solution.

innobead:
https://github.com/pierrec/lz4/blob/v4/internal/lz4block/decode_other.go which should be a fallback.

deltablock.go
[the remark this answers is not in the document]
innobead: Yeah, the corresponding engine/replicas will be gone.

derekbit:
> https://github.com/pierrec/lz4/blob/v4/internal/lz4block/decode_other.go which should be a fallback.

Interesting.
Original Lz4 is implementaed by C codes https://github.com/lz4/lz4.
The user (pierrec) wrote a ASM version. Is the binary compiled from GO is too slow?

innobead:
> > [pierrec/lz4@`v4`/internal/lz4block/decode_other.go](https://github.com/pierrec/lz4/blob/v4/internal/lz4block/decode_other.go?rgh-link-date=2023-01-10T14%3A00%3A19Z) which should be a fallback.
> 
> Interesting. Original Lz4 is implemented by C codes [lz4/lz4](https://github.com/lz4/lz4). The user (pierrec) wrote a ASM version. Is the binary compiled from GO too slow?

yeah, not sure if this is related to cgo overhead concern or cross-compile, so use asm instead. https://github.com/golang/go/issues/42469#issuecomment-753979705

You can also see go has systemcall for different arches, so this makes it can be cross-compiling for different arches.
image

Anyway, this is out of the topic, but it's good to know about lz4 package implementation :slightly_smiling_face:

PR #95, opened 2023-01-10
Refactor backupstore codes

derekbit:
longhorn/longhorn#5189

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -64,9 +64,7 @@ func addVolume(volume *Volume, driver BackupStoreDriver) error {
 		log.WithError(err).Errorf("Failed to add volume %v", volume.Name)
 		return err
 	}
-	log.Debug("Added backupstore volume ", volume.Name)
-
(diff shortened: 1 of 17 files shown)


backupstore.go
```diff
@@ -66,7 +66,7 @@ func addVolume(volume *Volume, driver BackupStoreDriver) error {
 	}
 	log.Debug("Added backupstore volume ", volume.Name)
```
shuo-wu, line 67: Redundant log?

shuo-wu (approved):
In general LGTM

backupstore.go
derekbit: Yeah. Rmoved. Thank you.

joshimoo (approved):
LGTM, thanks for pulling the refactoring changes out into their own PR

innobead (commented):
This is another good example to refacotr when available. Take any chance if possible.

PR #96, opened 2023-01-11
CIFS Backup Store Support

derekbit:
https://github.com/longhorn/longhorn/issues/3599


cifs/cifs.go
```diff
@@ -0,0 +1,122 @@
...
+		cmd = append(cmd, "-o", fmt.Sprintf("password=%v", b.password))
+	}
+	cmd = append(cmd, "//"+b.serverPath, b.mountDir)
+
+	_, err = util.ExecuteWithCustomTimeout("mount", cmd, defaultTimeout)
+	if err != nil {
+		return errors.Wrapf(err, "failed to mount using CIFS")
+	}
+	return nil
+}
```
innobead, line 110: It's time to use the library. Try https://pkg.go.dev/k8s.io/mount-utils? (even though`mount-utils` is still using exec to run the system commands, but at least can leverage some existing efforts)

innobead (changes requested):

cifs/cifs.go
derekbit: Looks good. Will change `nfs` and `cifs` together.

derekbit: mount timeout is added `https://github.com/longhorn/backupstore/commit/2b454435af5e09ada3fa01b316b60803160b2f2d` for Harvester.

https://pkg.go.dev/k8s.io/mount-utils doesn't not support context for cancelling the stuck mount operation. Should we use `mount-utils` or keep the current method?

cifs/cifs.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	cmd := []string{
+		"-t", "cifs",
+	}
+	if b.username != "" {
+		cmd = append(cmd, "-o", fmt.Sprintf("username=%v", b.username))
+	}
+	if b.password != "" {
+		cmd = append(cmd, "-o", fmt.Sprintf("password=%v", b.password))
```
l-mb, line 101: This exposes the login credentials via the process list.

At the very least, please pass them via the USER/PASSWD environment variable or the credentials file parameter.

cifs/cifs.go
derekbit: > Actually, this seems to imply there's no umount anywhere right now?

Yes

> An alternative that's easier to recover errors from is to not mount via the kernel, but use either go bindings for libsamba or run individual smbclient commands that don't touch the kernel state.

IIUC, it means we need to implement a set of file operations for the `smbclient`. Right? 

Ref: The conext of the cancellation of a mount is from this issue https://github.com/longhorn/backupstore/pull/91.

cifs/cifs.go
derekbit: Addressed. Thank you.

cifs/cifs.go
l-mb: > > Actually, this seems to imply there's no umount anywhere right now?
> Yes

Could this interfere with reboots maybe? It could definitely lead to hangs until the umount times out on reboot.

> > An alternative that's easier to recover errors from is to not mount via the kernel, but use either go bindings for libsamba or run individual smbclient commands that don't touch the kernel state.
> IIUC, it means we need to implement a set of file operations for the `smbclient`. Right?

Right, this would be similar to how S3 is implemented.

> Current, except for the s3 protocol, vfs, nfs and cifs rely on the generic filesystem operations. The hanging mount rarely happened before. I think we can still use the `mount` here. In the future, we can improve it by using libsamba or smbclient instead if needed.

Yes, NFS is similar. It's rare, but has advantages in terms of robustness and kernel dependency reduction.

> Ref: The conext of the cancellation of a mount is from this issue #91.

That doesn't really cancel a mount, it merely kills the process and hopes it catches it at a time where that then indeed cleanly aborts the mount. That can be problematic if it doesn't, and then a potential retry later might not succeed, or the kernel might have dangling mounts somewhere.

`umount -f` exists, but is not as reliable.

You're right though that this is an easier approach since VFS/NFS are already implemented in a similar fashion.

cifs/cifs.go
```diff
@@ -0,0 +1,122 @@
...
+	username string
+	password string
+
+	*fsops.FileSystemOperator
+}
+
+const (
+	KIND = "cifs"
+
+	MountDir = "/var/lib/longhorn-backupstore-mounts"
```
shuo-wu, line 41: NIT: should we use a dir that is different from the NFS part?

cifs/cifs.go
shuo-wu: FYI, there is already an issue related to the stuck umount. And we mentioned in the ticket, there is no timeout/cleanup mechanism in the Kubernetes `mount-utils`:
https://github.com/longhorn/longhorn/issues/5296

derekbit:
> FYI, there is already an issue related to the stuck umount. And we mentioned in the ticket, there is no timeout/cleanup mechanism in the Kubernetes mount-utils:
https://github.com/longhorn/longhorn/issues/5296

Than you. Yeah, I've already handled this issue in the lastes update. You can check it again.

cifs/cifs.go
derekbit: Because we only support one backup store, so I think using the same dir is enough.

cifs/cifs.go
derekbit: Right. As the solution in longhorn/longhorn#5296, the force unmount is added.

util/util.go
```diff
@@ -243,3 +250,50 @@ func IsMounted(mountPoint string) bool {
...
+	log.Infof("Trying to clean up mount point %v", mountDir)
+	return mount.CleanupMountPoint(mountDir, forceUnmounter, false)
+}
+
+func CheckAndCleanupMountPoint(Kind, mountPoint string, mounter mount.Interface, log logrus.FieldLogger) (bool, error) {
+	mounted, err := mounter.IsMountPoint(mountPoint)
+	if err != nil {
+		if !strings.Contains(err.Error(), syscall.ENOENT.Error()) {
+			return false, nil
+		}
```
shuo-wu, line 270: Is this mean `No such file or directory`? If YES, we can put `mkdir` here, which is like what func `ensureMountPoint` in longhorn-manager does.

util/util.go
```diff
@@ -243,3 +250,50 @@ func IsMounted(mountPoint string) bool {
...
+		return true, errors.Wrapf(err, "failed to get mount for %v", mountPoint)
+	}
+
+	if strings.Contains(mnt.FilesystemType, Kind) {
+		_, statErr := os.Stat(mountPoint)
+		if statErr == nil {
+			return true, nil
+		}
+		log.WithError(statErr).Warnf("Failed to get file info of mount point %v", mountPoint)
+	}
```
shuo-wu, line 292: Should we check if it's a valid mount point by reading the dir? See func `ensureMountPoint` in longhorn-manager

nfs/nfs.go
```diff
@@ -107,6 +107,9 @@ func (b *BackupStoreDriver) mount() (err error) {
 		mountOptions := []string{
 			fmt.Sprintf("nfsvers=%v", version),
 			"actimeo=1",
+			"soft",
```
shuo-wu, line 110: Compared with the RWX volume part, we don't need the strict retry/recover mechanism hence it's fine to use soft mode, right?

util/util.go
derekbit: Renamed the function to `EnsureMountPoint` and added the `MkdirAll` as well. Thank you.

util/util.go
derekbit: Addressed. Thank you.

cifs/cifs.go
```diff
@@ -65,48 +68,57 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
+
+	mountComplete := false
+	err := wait.PollImmediate(1*time.Second, defaultTimeout, func() (bool, error) {
+		err := mounter.MountSensitiveWithoutSystemd("//"+b.serverPath, b.mountDir, KIND, mountOptions, sensitiveMountOptions)
+		mountComplete = true
+		return true, err
+	})
+	if !mountComplete {
+		return errors.Wrapf(err, "mounting CIFS share %v on %v timed out", b.destURL, b.mountDir)
 	}
```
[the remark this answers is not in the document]
derekbit: Added `MountWithTimeout` for the abstraction. Thank you.

nfs/nfs.go
derekbit: Yes. For backup, we can simply fail the backup and avoid the client getting stuck forever if the backup store is somehow broken.

util/util.go
```diff
@@ -243,3 +250,71 @@ func IsMounted(mountPoint string) bool {
...
+	mnt, err := fs.GetMount(mountPoint)
+	if err != nil {
+		return true, errors.Wrapf(err, "failed to get mount for %v", mountPoint)
+	}
+
+	if strings.Contains(mnt.FilesystemType, Kind) {
+		_, statErr := os.Stat(mountPoint)
+		if statErr == nil {
+			return true, nil
+		}
```
shuo-wu, line 312: NIT: Do we still need this?

shuo-wu (approved):
In general LGTM

util/util.go
derekbit: No, we don't need it. Removed. Thank you.

shuo-wu (approved):

util/util.go
```diff
@@ -243,3 +250,69 @@ func IsMounted(mountPoint string) bool {
...
+		if strings.Contains(err.Error(), syscall.ENOENT.Error()) {
+			return false, nil
+		}
+	}
+
+	IsCorruptedMnt := mount.IsCorruptedMnt(err)
+	if !IsCorruptedMnt {
+		logrus.Warnf("Mount point %v is trying reading dir to make sure it is healthy", mountPoint)
+		if _, err := ioutil.ReadDir(mountPoint); err != nil {
+			logrus.Warnf("Mount point %v was identified as corrupt by ReadDir", mountPoint)
```
innobead, line 286: Should include `err` in the log.

util/util.go
```diff
@@ -243,3 +250,69 @@ func IsMounted(mountPoint string) bool {
...
+			IsCorruptedMnt = true
+		}
+	}
+
+	if IsCorruptedMnt {
+		log.WithError(err).Warnf("Failed to check mount point %v (mounted=%v)", mountPoint, mounted)
```
innobead, line 292: Where is the value of `err` from? Probably no need to include `err` in this log because `err` should be handled already.

util/util.go
```diff
@@ -243,3 +250,69 @@ func IsMounted(mountPoint string) bool {
...
+		if mntErr := cleanupMount(mountPoint, mounter, log); mntErr != nil {
+			return true, errors.Wrapf(err, "failed to clean up corrupted mount point %v", mountPoint)
```
innobead, line 294: Should wrap `mntErr` instead? or just replace `mntErr` with a local `err`?

cifs/cifs.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/longhorn/backupstore"
+	"github.com/longhorn/backupstore/fsops"
+	"github.com/longhorn/backupstore/util"
+)
+
+var (
+	log                  = logrus.WithFields(logrus.Fields{"pkg": "cifs"})
+	defaultMountInterval = 1 * time.Second
+	defaultMountTimeout  = 5 * time.Second
```
innobead, line 24: note: exact meaning for the const is good.

innobead (changes requested):
In general, LGTM. Just some minor things needed to fix.

util/util.go
derekbit: Sorry, forgot removing `err`.

util/util.go
derekbit: Fixed. Thank you.

util/util.go
derekbit: Fixed. Thank you.

util/util.go
derekbit: Fixed. Thank you.

cifs/cifs.go
derekbit: Added the relevant link for reference. Thank you.

innobead: I mean the variable rename is good and meaningful, so you actually don't need to do anything. This is just a note 🙂

derekbit: Got it. I still prefer the link here, because the 5-second timeout is dedicated for the downstream (harvester project). For other projects and developers, the short value might be confusing.

innobead (approved):

PR #97, opened 2023-01-12
fix(http): proxy setting could not be changed

mantissahz:
Proxy setting could not be changed by http.ProxyFromEnvironment after grpc server is running up.
Set our proxy function to get the proxy URL when creating a new http client for initializing s3 driver.

longhorn/longhorn#5054


shuo-wu:
> Proxy setting could not be changed by http.ProxyFromEnvironment after grpc server is running up.

Can you talk a bit more about this?

mantissahz:
>>Proxy setting could not be changed by http.ProxyFromEnvironment after grpc server is running up.

>Can you talk a bit more about this?

@shuo-wu 
Sure, 
`http.ProxyFromEnvironment` is the default setting of `Transport` to get the environment Proxy setting. https://github.com/golang/go/blob/master/src/net/http/transport.go#L43-L54
When IM-e GPRC server is running and starts to communicate with `longhorn-manager`, `Transport.Proxy` will be run only once to get the proxy setting [envProxyOnce.Do](https://github.com/golang/go/blob/master/src/net/http/transport.go#L827-L832).
Changing/Setting environment variables `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` in the GRPC server when getting the requests will not affect the `http.ProxyFromEnvironment` even creating a new client in the GRPC server

http/client.go
```diff
@@ -44,5 +46,48 @@ func GetClient(insecure bool, customCerts []byte) (*http.Client, error) {
 		RootCAs:            certs,
 	}
 	client := &http.Client{Transport: customTransport}
+	customTransport.Proxy = proxyForURL
```
innobead, line 49: Why not just? Also, not just these 2 environment variables, also need to consider `NO_PROXY`, so it's better to just leverage the original lib directly. 

```go
	customTransport := http.DefaultTransport.(*http.Transport).Clone()
    customTransport.Proxy = func(request *http.Request) (*url.URL, error) {
        return httpproxy.FromEnvironment().ProxyFunc()(request.URL)
    }
    customTransport.TLSClientConfig = &tls.Config{
        InsecureSkipVerify: insecure,
        RootCAs:            certs,
    }
    client := &http.Client{Transport: customTransport}
```

innobead (changes requested):

http/client.go
mantissahz: Definitely a better way. Modified and Thanks.

innobead (approved):

mantissahz:
Start to update longhorn-instance-manager

derekbit:
Can we create a v1.4.x branch in backupstore?
longhorn-engine still uses older commit without the backup compression improvement, but longhorn-instance-manager needs a newer one containing this commit. The difference make the dependency update a bit messy.
https://github.com/longhorn/backupstore/commits/master

cc @innobead @mantissahz

innobead:
It seems inevitable but it is definitely anti pattern. That means we probably also need for 1.3.x because it's still under maintenance.

PR #98, opened 2023-01-16
Fall back compression/decompression method to gzip for existing backups

derekbit:
Longhorn/longhorn#5189


innobead (approved):
LGTM

PR #99, opened 2023-01-19
fix(http): set Proxy function for all http client of s3 driver

mantissahz:
Use `GetClient` to create the http client for s3 driver.

longhorn/longhorn#5054

Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2937/

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -61,13 +61,12 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 	}
 
 	// add custom ca to http client that is used by s3 service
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):

joshimoo (approved):
LGTM

Good change, we will now always use a custom HTTP client instead of the default http client used by AWS sdk if non is present.

Since the proxy setup was added to our custom http setup but not the default client, this addresses the issue for all clusters including the ones that don't require a custom CA certificate.

innobead:
@mantissahz Please remember to update the corresponding repo dependencies.

PR #101, opened 2023-02-14
Return error if mount failed

derekbit:
Longhorn/longhorn#3599

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -329,5 +329,5 @@ func MountWithTimeout(mounter mount.Interface, source string, target string, fst
 	if !mountComplete {
 		return errors.Wrapf(err, "mounting %v share %v on %v timed out", fstype, source, target)
 	}
-	return nil
+	return err
 }


util/util.go
```diff
@@ -329,5 +329,5 @@ func MountWithTimeout(mounter mount.Interface, source string, target string, fst
 	if !mountComplete {
 		return errors.Wrapf(err, "mounting %v share %v on %v timed out", fstype, source, target)
 	}
-	return nil
+	return err
```
shuo-wu, line 332: Can there be a case that `mountComplete` is true but the error is not empty?

derekbit: Yeah, for example, giving a wrong address.

shuo-wu: Does this mean that the device will be mounted at the wrong address? If YES, does this function need to revert the change (do unmount) when there is an error?

derekbit: > Does this mean that the device will be mounted at the wrong address?

I mean a not existing remote address. The error will be returned.

But you're right. For a existing but wrong address, we should unmount it. I will open another ticket for this issue.

shuo-wu (approved):

innobead (approved):

innobead:
@Mergifyio backport v1.4.x v1.3.x

innobead:
https://github.com/Mergifyio backport v1.4.x v1.3.x

PR #102, opened 2023-02-16
Clean up unused mount points

derekbit:
Longhorn/longhorn#5391


util/util.go
```diff
@@ -331,3 +335,33 @@ func MountWithTimeout(mounter mount.Interface, source string, target string, fst
...
+		if err != nil {
+			errs = multierr.Append(errs, errors.Wrapf(err, "failed to check if %s is mounted", path))
+			return nil
+		}
+
+		if mounted {
+			if err := cleanupMount(path, mounter, log); err != nil {
+				errs = multierr.Append(errs, errors.Wrapf(err, "failed to clean up mount point %v", path))
+			}
+		}
```
innobead, line 362: This part needs to update due to the `IsMountPoint` is incompatible with the version used by longhorn-manager.

util/util.go
```diff
@@ -331,3 +335,33 @@ func MountWithTimeout(mounter mount.Interface, source string, target string, fst
 	}
 	return err
 }
+
+func CleanUpMountPoints(mounter mount.Interface, log logrus.FieldLogger) error {
```
innobead, line 339: Add a doc for this method.

innobead (approved):
In general, LGTM.

However, this means this library will only support one mount point at once. Suggest creating an improvement issue to make the library support multiple mounts capability for the future. (ex: support mutliple backups to different backup stores)

innobead (changes requested):
Forgot some comments that need to resolve first.

util/util.go
derekbit: Hmm, where is the doc?

innobead: Something like

```go
// CleanUpMountPoints will try to clean up all existing mount points for existing backup stores
func CleanUpMountPoints(mounter mount.Interface, log logrus.FieldLogger) error {
...
```

util/util.go
derekbit: Add a feature request https://github.com/longhorn/longhorn/issues/5411.
Thank you.

util/util.go
derekbit: Added. Thank you.

innobead (approved):
LGTM. :+1: ! 

Awesome to add docs for other exported functions :slightly_smiling_face:

PR #103, opened 2023-02-17
vendor: force use v0.23.6 apimachinery and mount-utils

derekbit:
Align the k8s packages because longhorn-manager force uses v0.23.6.

Longhorn/longhorn#3599

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -329,5 +329,5 @@ func MountWithTimeout(mounter mount.Interface, source string, target string, fst
 	if !mountComplete {
 		return errors.Wrapf(err, "mounting %v share %v on %v timed out", fstype, source, target)
 	}
-	return nil
+	return err
 }


innobead:
@derekbit Will this change impact 1.3 and 1.4 following backports relying on the backupstore lib? I assume NO, because relying on the mount until just started from CIFS introduction.

derekbit:
Just noticed that the k8s packages use v0.23.6 (https://github.com/longhorn/longhorn-manager/blob/master/go.mod#L28).
To avoid the API compability error, the mount-utils and apimachinary should use v0.23.6 as well.

derekbit:
> @derekbit Will this change impact 1.3 and 1.4 following backports relying on the backupstore lib? I assume NO, because relying on the mount until just started from CIFS introduction.

No, mount-utils and apimachinary are firstly used in backupstore from this CIFS feature.

PR #107, opened 2023-03-06
Backport proxy setting to v1.4.x branch

derekbit:
Longhorn/longhorn#5054


innobead (approved):

innobead:
The goal is to ensure the backupstore used by IM and longhorn-engine is the same version.

PR #108, opened 2023-03-06
vendor: update go to 1.17

derekbit:


innobead:
longhorn-engine 1.4.x seems still 1.13, so need to update it as well; otherwise, the change here will be superseded by the dependant.

derekbit:
Ready for review

innobead (approved):

PR #113, opened 2023-03-24
Fix wrong mounter initialization

derekbit:
The parameter mounterPath of mount.New(...) is the path to directory contianing a customized mount command. Should leave it as empty for using the system mount command.

Longhorn/longhorn#3599


innobead (approved):
LGTM

PR #114, opened 2023-03-24
[BACKPORT][v1.3.3] Refactor NFS backupstore and use soft mode for NFS instead

derekbit:
Refactor NFS backstore
Usingsoft mode for NFS instead
Check and clean up corrupted mount point before backing up data

https://github.com/longhorn/longhorn/issues/5628
https://github.com/longhorn/longhorn/issues/5654


innobead:
@derekbit is this ready for review?

derekbit:
> @derekbit is this ready for review?

Yeah. Updated go to v1.17. Please check if there is any concern for LH v1.3.x. Thank you.

innobead:
This really depends on the dependent components, in 1.3, all are based on 1.3, so we probably need some further testing before merge.

derekbit:
> This really depends on the dependent components, in 1.3, all are based on 1.3, so we probably need some further testing before merge.

I see. Will run e2e first.

innobead:
@derekbit Probably no need if all dependent components can be built successfully, because go is compiled language.

derekbit:
> @derekbit Probably no need if all dependent components can be built successfully, because go is compiled language. However, it's good to see if backup related test cases work fine.

Update go of longhorn-engine and longhorn-instance-manager to v1.17 as well. Everything works well.
https://github.com/longhorn/longhorn-engine/pull/879
https://github.com/longhorn/longhorn-instance-manager/pull/203

derekbit:
Won't apply to v1.3.x because the feature will change the backup behavior.

PR #117, opened 2023-03-28
backup: introduce "starting" state

derekbit:
To make the backup state machine more clear, this commit introduces a new "starting" state to the existing "in_progress," "complete," and "error" states.

Previously, a backup task consisted of preparation, backupstore initialization, and backup execution, with "in_progress" starting from backup execution. However, this meant that any stuck caused by an NFS connection during backupstore initialization could not be detected by the Longhorn manager. With the new "starting" state, the backup process will have a more complete state machine, ensuring better tracking and monitoring of the backup process.

Longhorn/longhorn#5662


util/util.go
```diff
@@ -379,3 +379,10 @@ func CleanUpMountPoints(mounter mount.Interface, log logrus.FieldLogger) error {
 
 	return errs
 }
+
+func GetBackupNameFromBackupNameInConfig(backupNameInConfig string) string {
```
shuo-wu, line 383: Is this necessary (in `util/util`)? Now it's used by the test only.

deltablock.go
```diff
@@ -236,19 +236,21 @@ func CreateDeltaBlockBackup(config *DeltaBackupConfig) (backupID string, isIncre
...
 	if err := lock.Lock(); err != nil {
 		deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
-		return "", backupRequest.isIncrementalBackup(), err
+		return backupRequest.isIncrementalBackup(), err
 	}
 	go func() {
 		defer deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
 		defer lock.Unlock()
 
+		deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(ProgressStateInProgress), 0, "", "")
```
shuo-wu, line 245: The main purpose of adding an input `state` for this function is, we need to update the state to `in_progress` with `0` progress here, right?

derekbit: Yes. In addition, we can set the state to `Starting`.

util/util.go
derekbit: Moved into test. Thank you.

deltablock.go
```diff
@@ -106,48 +106,53 @@ const (
 	PROGRESS_PERCENTAGE_BACKUP_TOTAL    = 100
 )
 
-func CreateDeltaBlockBackup(config *DeltaBackupConfig) (backupID string, isIncremental bool, err error) {
+func CreateDeltaBlockBackup(backupName string, config *DeltaBackupConfig) (isIncremental bool, err error) {
 	if config == nil {
-		return "", false, fmt.Errorf("invalid empty config for backup")
+		return false, fmt.Errorf("BUG: invalid empty config for backup")
```
innobead, line 111: nit: `BUG:` is not quite meaningful, even though we know it has been used in some logs to highlight unexpected error. Ideally, we should use a specific error structure instead. (let's review this in general in the future)

ref: https://github.com/longhorn/longhorn/issues/5837

deltablock.go
```diff
@@ -106,48 +106,53 @@ const (
...
 	destURL := config.DestURL
 	deltaOps := config.DeltaOps
 	if deltaOps == nil {
-		return "", false, fmt.Errorf("missing DeltaBlockBackupOperations")
+		return false, fmt.Errorf("BUG: missing DeltaBlockBackupOperations")
 	}
 
+	defer func() {
+		if err != nil {
+			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(ProgressStateError), 0, "", err.Error())
```
innobead, line 123: so, I would expect when encountering any error during delta backup, the progress will be reset to 0? I am curious why not just keeping the original progress?

deltablock.go
```diff
@@ -86,7 +86,7 @@ type DeltaBlockBackupOperations interface {
 	OpenSnapshot(id, volumeID string) error
 	ReadSnapshot(id, volumeID string, start int64, data []byte) error
 	CloseSnapshot(id, volumeID string) error
-	UpdateBackupStatus(id, volumeID string, backupProgress int, backupURL string, err string) error
+	UpdateBackupStatus(id, volumeID string, backupState string, backupProgress int, backupURL string, err string) error
```
innobead, line 89: Ideally, we should not change the spec of any exported interface but rather create another interface method instead.

However, in this case, `UpdateBackupStatus` seems not being used by any caller directly, so it's fine here.

innobead (approved):
In genreal, LGTM.

Unfortunately, we have 1.4.x & 1.3.x branch for backupstore, so we need to backport. (We should keep the master backward compatibilty for libraries.)

deltablock.go
derekbit: I agree with your opinion, but the progress is always set to 0 when encountering an error([code](https://github.com/longhorn/longhorn-engine/blob/master/pkg/replica/backup.go#L176)). Do you suggest we keep the progress?

deltablock.go
derekbit: Yeah, we have many `BUG: ....` in longhorn components.

image

deltablock.go
derekbit: I think it is feasible. Is there any advantage?

deltablock.go
innobead: I would say yes, and we will know what progress/state it failed at, even though we have an error message as well. This is minor, we can do some improvements when tackling https://github.com/longhorn/longhorn/issues/5837.

deltablock.go
innobead: I think what @shuo-wu said is if we can update the status directly, then we don't need to introduce an extra starting state.

innobead: then, let's do this way to make it compatible.

innobead:
@Mergifyio backport v1.4.x

derekbit:
@Mergify v1.4.x

derekbit:
@mergify backport v1.4.x

PR #118, opened 2023-04-11
feat: add new command to unmount all mount points

ChanYiLin:
ref: [longhorn/longhorn#5655](https://github.com/longhorn/longhorn/issues/5655)

- add a new command for cleaning up all mount points
- manager will call this command when backup target is unset


innobead:
@Mergifyio backport v1.4.x

PR #120, opened 2023-04-18
feat: Azure Blob Storage Support

mantissahz:
Introduce Azure blob storage support.

Ref: longhorn/longhorn#1309


azblob/azblob_service.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return blobs, nil
+}
+
+func (s *Service) GetBlobProperties(blob string) (azblob.GetBlobPropertiesResponse, error) {
+	container, err := s.New()
+	if err != nil {
+		return azblob.GetBlobPropertiesResponse{}, err
+	}
+	blobClient := container.NewBlockBlobClient(blob)
```
derekbit, line 75: nil: abstract this part to a function

```suggestion
func (s *Service) newBlockBlobClient() ... {
	....
}
```

azblob/azblob_service.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+	defer s.Close()
+
+	var deletionFailures []string
+	for _, blob := range blobs {
+		blobClient := container.NewBlockBlobClient(blob)
+		_, err = blobClient.Delete(context.Background(), nil)
+		if err != nil {
+			log.Errorf("Failed to blob object: %v error: %v", blob, err)
```
derekbit, line 135:
```suggestion
			log.WithError(err).Errorf("Failed to blob object: %v", blob)
```

azblob/azblob_service.go
mantissahz: Addressed. Thanks

azblob/azblob_service.go
mantissahz: Modified. Thanks

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+	AccountName string
+	AccountKey  string
+}
+
+func (s *Service) New() (azblob.ContainerClient, error) {
+	cred, err := azblob.NewSharedKeyCredential(s.AccountName, s.AccountKey)
+	if err != nil {
+		return azblob.ContainerClient{}, err
+	}
+	url := fmt.Sprintf(azureServiceURL, s.AccountName)
```
innobead, line 29: Should we need to hard code `azureServiceURL`? Is there any chance to use a customized endpoint? We support using a custom endpoint for s3.

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+	pager := container.ListBlobsHierarchy(delimiter, listOptions)
+
+	for pager.NextPage(context.Background()) {
+		resp := pager.PageResponse()
+		for _, v := range resp.ContainerListBlobHierarchySegmentResult.Segment.BlobItems {
+			blobs = append(blobs, *v.Name)
+		}
+		for _, v := range resp.ContainerListBlobHierarchySegmentResult.Segment.BlobPrefixes {
+			blobs = append(blobs, *v.Name)
+		}
```
innobead, line 60: Why need to collect `BlobItem` and `BlobPrefix` together? What is `BlobPrefix`? I assume what we want is `BlobItem`?

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if err := pager.Err(); err != nil {
+		return []string{}, err
+	}
+
+	return blobs, nil
+}
+
+func (s *Service) GetBlobProperties(blob string) (azblob.GetBlobPropertiesResponse, error) {
```
innobead, line 70: Change the returned variables to `*azblob.GetBlobPropertiesResponse, error`. This should apply to all methods if they will return error.

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+	blobClient, err := s.newBlockBlobClient(blob)
+	if err != nil {
+		return azblob.GetBlobPropertiesResponse{}, err
```
innobead, line 73: `return nil, err`. The same change is for other methods after changing the returned values.

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+
+	container, err := s.New()
+	if err != nil {
+		return err
+	}
+	defer s.Close()
+
+	var deletionFailures []string
+	for _, blob := range blobs {
+		blobClient := container.NewBlockBlobClient(blob)
```
innobead, line 137: Why not just have one client for the service, then reused in all method calls? I know you learn from s3, but the implementation there is not good.

In this implementation, in each method call, you will create a client again and this would cause extra unnecessary cost.

innobead (changes requested):
Please fix all comments. Thanks.

innobead:
also, PR check failed.

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/pkg/errors"
+)
+
+const (
+	azureServiceURL = "https://%s.blob.core.windows.net/"
+)
+
+type Service struct {
```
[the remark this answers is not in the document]
mantissahz: `Service` is exported by the driver initialization now.
And removed `New` function by `NewService`.

azblob/azblob.go
```diff
@@ -0,0 +1,192 @@
...
+func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
+	b := &BackupStoreDriver{}
+
+	u, err := url.Parse(destURL)
+	if err != nil {
+		return nil, err
+	}
+
+	if u.Scheme != KIND {
+		return nil, fmt.Errorf("BUG: Why dispatch %v to %v?", u.Scheme, KIND)
```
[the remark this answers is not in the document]
mantissahz: Modified the error message.

azblob/azblob.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	if u.User != nil {
+		b.service.ServiceURL = u.Host
+		b.service.Container = u.User.Username()
+	} else {
+		b.service.Container = u.Host
+	}
```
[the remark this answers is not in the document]
mantissahz: Removed and used `NewService` in azblob_service.go instead

azblob/azblob_service.go
mantissahz: Used `AZBLOB_ENDPOINT` and `s.ServiceURL` parsed from `backup-target` URL instead.

azblob/azblob_service.go
mantissahz: Modified. Thanks.

azblob/azblob_service.go
mantissahz: Because `backupstore/volumes/02/a8/` will be only added into `BlobPrefix` and when [listing backup volumes on remote backup target](https://github.com/longhorn/backupstore/blob/master/config.go#L151-L242), it needs to list the lv1Dir `backupstore/volumes/02/` and lv2Dir `backupstore/volumes/02/a8/`.

azblob/azblob_service.go
mantissahz: Modified. Thanks.

azblob/azblob_service.go
mantissahz: Need to get blob client for each blob but container client is modified to be initialized once.

azblob/azblob.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (s *BackupStoreDriver) FileExists(filePath string) bool {
+	return s.FileSize(filePath) >= 0
+}
+
+func (s *BackupStoreDriver) FileSize(filePath string) int64 {
+	path := s.updatePath(filePath)
+	head, err := s.service.GetBlobProperties(path)
+	if err != nil || head.ContentLength == nil {
```
innobead, line 120: Don't consume `err` silently. Need to log error.

azblob/azblob.go
```diff
@@ -0,0 +1,187 @@
...
+		return -1
+	}
+	return *head.ContentLength
+}
+
+func (s *BackupStoreDriver) FileTime(filePath string) time.Time {
+	path := s.updatePath(filePath)
+	blobProp, err := s.service.GetBlobProperties(path)
+	if err != nil || blobProp.ContentLength == nil {
```
innobead, line 129: The same, don't silently consume `err`.

azblob/azblob.go
```diff
@@ -0,0 +1,187 @@
...
+	return filepath.Join(s.path, path)
+}
+
+func (s *BackupStoreDriver) List(listPath string) ([]string, error) {
+	var result []string
+
+	path := s.updatePath(listPath) + "/"
+	contents, err := s.service.ListBlobs(path, "/")
+	if err != nil {
+		log.WithError(err).Error("Failed to list azblob")
```
innobead, line 92: No need to have this, because the error will be returned back to the caller.

azblob/azblob_service.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	s.ContainerClient = serviceClient.NewContainerClient(s.Container)
+	return &s, nil
+}
+
+func (s *Service) Close() {
+}
+
+func (s *Service) ListBlobs(prefix, delimiter string) (*[]string, error) {
```
innobead, line 69: Add a comment to this method to descirbe the returned items could be blobs or blob prefixes.

mantissahz: Added the comment to this method and other exported methods. Thanks

azblob/azblob.go
mantissahz: Added an error log. Thanks

azblob/azblob.go
mantissahz: Added an error log. Thanks

azblob/azblob.go
mantissahz: Removed. Thanks

azblob/azblob_service.go
```diff
@@ -0,0 +1,150 @@
...
+
+	s.ContainerClient = serviceClient.NewContainerClient(s.Container)
+	return &s, nil
+}
+
+func (s *service) Close() {
+}
+
+// ListBlobs returns items that contains blobs or blob prefixes
+func (s *service) ListBlobs(prefix, delimiter string) (*[]string, error) {
```
innobead, line 70: as I mentioned, don't need to export the methods of service structure. Also, is `Close()` required?

innobead (changes requested):

azblob/azblob_service.go
mantissahz: Modified all methods to be not exported and removed method `Close()`, yes it is not required.
Thanks

innobead (approved):

innobead:
@mantissahz the below code seems incorrect. We should always override `s.ServiceURL` if `azureEndpoint` is not empty, right?

[/azblob/azblob_service.go#L44-L55](https://github.com/longhorn/backupstore/blob/6183661a2bd13517f6c683d8c06fb9dd7d5baca1/azblob/azblob_service.go#L44-L55)
```go
	connStr := fmt.Sprintf(azureConnNameKey, accountName, accountKey)
	if azureEndpoint != "" {
		blobEndpointURL := strings.TrimRight(azureEndpoint, "/") + "/" + accountName
		connStr = fmt.Sprintf(connStr+blobEndpoint, blobEndpointURL)
	}

	if s.ServiceURL == "" {
		s.ServiceURL = azureURL
	}
	if s.ServiceURL != azureURL {
		connStr = connStr + fmt.Sprintf(blobEndpointSuffix, s.ServiceURL)
	}

```

mantissahz:
`azureEndpoint` will use `blobEndpoint` (`"BlobEndpoint=%s;"`)
and `s.ServiceURL` will use `blobEndpointSuffix` (`"EndpointSuffix=%s;"`) (Default `EndpointSuffix` is `azureURL `)
They will be concatenated together into `connStr`.

innobead:
You are right. It will be always respected if the value is provided.

PR #121, opened 2023-04-18
fix: replace jobq with workerpool lib to fix license issue

ChanYiLin:
ref: longhorn/longhorn 4176

Before we used `honestbee/jobq` for concurrently getting all volumes from backupstore. We now have to replace it with `gammazero/workerpool` since `honestbee/jobq` has a LICENSE issue.

`honestbee/jobq` provides following features
- Create workerpool
- Submit a task to workerpool with a timeout setting, and get a `TaskTracker` as a return value for getting the task results when it is done.

`gammazero/workerpool` on the other hand,
- Has no `TaskTracker` for tracking the result of the task 
    - => We have to create a channel to get the results from the task.
- Cannot set timeout for the task 
    - => However, we already have a timeout for `util.Execute`. Thus, the command will timeout if it can't finish in 1 min(default value).


ChanYiLin:
e2e test passed, run every test 3 times.
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3508/testReport/

list.go
```diff
@@ -90,18 +90,21 @@ func List(volumeName, destURL string, volumeOnly bool) (map[string]*VolumeInfo,
...
+	jobQueues := workerpool.New(runtime.NumCPU() * 16)
+	defer jobQueues.StopWait()
+
+	// jobQueues := jobq.NewWorkerDispatcher(
+	// 	// init #cpu*16 workers
+	// 	jobq.WorkerN(runtime.NumCPU()*16),
+	// 	// init worker pool size to 256 (same as max folders 16*16)
+	// 	jobq.WorkerPoolSize(256),
+	// )
+	// defer jobQueues.Stop()
```
innobead, line 102: Please remove these comments.

config.go
```diff
@@ -154,52 +152,56 @@ func getVolumeNames(jobQueues *jobq.WorkerDispatcher, jobQueueTimeout time.Durat
...
+	lv1Trackers := make(chan JobResult)
+	lv2Trackers := make(chan JobResult)
+	defer close(lv1Trackers)
+	defer close(lv2Trackers)
+
 	for _, lv1Dir := range lv1Dirs {
 		path := filepath.Join(volumePathBase, lv1Dir)
-		lv1Tracker := jobQueues.QueueTimedFunc(context.Background(), func(ctx context.Context) (interface{}, error) {
+		jobQueues.Submit(func() {
 			lv2Dirs, err := driver.List(path)
```
innobead, line 164: Ideally, we should prevent any chance to let submitted tasks running forever if they are expected.

Relying on how the caller to determine the timeout is not an ideal way since it is uncertain.

Basically, you are still able to control this to make the submitted closure to run in a timeout wrapper closure via https://github.com/slok/goresilience/blob/master/examples/timeout/timeout.go. I would suggest doing this way.

innobead (changes requested):
In general, LGTM.

Only thing is we still need to introduce timeout to prevent any chances to make tasks running indefinitely.

innobead:
> e2e test passed, run every test 3 times. https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3508/testReport/

This is good!

config.go
ChanYiLin: Fixed, thanks

list.go
ChanYiLin: Removed

PR #129, opened 2023-05-04
Fix timeout caused by dead mount points

derekbit:
https://github.com/longhorn/longhorn/issues/5856


util/util.go
```diff
@@ -48,6 +49,17 @@ type NopCloser struct {
...
+	switch fstype {
+	case unix.NFS_SUPER_MAGIC:
+		return "nfs", nil
+	case unix.CIFS_SUPER_MAGIC:
+		return "cifs", nil
+	default:
+		return "", fmt.Errorf("unknown fstype %v", fstype)
+	}
+}
+
```
derekbit, line 62: Remove `case unix.CIFS_SUPER_MAGIC:` for v1.4.x

innobead (approved):
LGTM

PR #131, opened 2023-05-05
Fix build failure on s390x

derekbit:
https://github.com/longhorn/longhorn/issues/5856

```
vendor/github.com/longhorn/backupstore/util/util.go:334:28: cannot use stat.Type (variable of type uint32) as int64 value in argument to fstypeToKind
```


derekbit:
@Mergify backport v1.4.x

innobead (approved):

PR #133, opened 2023-05-15
feat(backup): store storage class name to backup volume

ChanYiLin:
ref: [longhorn/longhorn#3506](https://github.com/longhorn/longhorn/issues/3506)


innobead (approved):
LGTM. 

Need to ensure it's compatible with the previously existing backup volume config which doesn't have storage class field.

PR #135, opened 2023-06-05
fix(system-backup): unexpected bulk deletion

c3y1huang:
https://github.com/longhorn/longhorn/issues/6045

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -105,13 +105,24 @@ func (s *BackupStoreDriver) GetURL() string {
 }
 
 func (s *BackupStoreDriver) updatePath(path string) string {
-	return filepath.Join(s.path, path)
+	joinedPath := filepath.Join(s.path, path)
+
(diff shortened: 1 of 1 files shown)


s3/s3.go
```diff
@@ -105,13 +105,22 @@ func (s *BackupStoreDriver) GetURL() string {
 }
 
 func (s *BackupStoreDriver) updatePath(path string) string {
-	return filepath.Join(s.path, path)
+	joinedPath := filepath.Join(s.path, path)
+
+	if !strings.HasSuffix(path, "/") {
```
innobead, line 110: Suggest adding a comment to explain why do this way.

innobead (changes requested):
LGTM, but need a comment.

innobead (approved):

innobead:
@c3y1huang work with QA to check if this issue also happened to other backup stores.

c3y1huang:
@mergify backport v1.4.x

PR #137, opened 2023-06-12
A backup target backed by a Samba server is not recognized

derekbit:
Longhorn/longhorn#6100


innobead (approved):
LGTM. Good check!

innobead:
cc @longhorn/qa

PR #138, opened 2023-06-20
chore: set appropriate levels for log messages

derekbit:


config.go
```diff
@@ -174,8 +174,8 @@ func getVolumeNames(jobQueues *workerpool.WorkerPool, driver BackupStoreDriver)
 			err := runner.Run(context.TODO(), func(_ context.Context) error {
 				lv2Dirs, err := driver.List(path)
 				if err != nil {
-					logrus.Warnf("Failed to list second level dirs for path %v", path)
-					return fmt.Errorf("Failed to list second level dirs for path %v: %v", path, err)
+					logrus.WithError(err).Warnf("Failed to list second level dirs for path %v", path)
```
innobead, line 177: The error will be returned back to the caller, so the caller should deal with the error like error out. A duplicated log is redundant?

config.go
```diff
@@ -209,8 +209,8 @@ func getVolumeNames(jobQueues *workerpool.WorkerPool, driver BackupStoreDriver)
 				err := runner.Run(context.TODO(), func(_ context.Context) error {
 					volumeNames, err = driver.List(path)
 					if err != nil {
-						logrus.Warnf("Failed to list volume names for path %v", path)
-						return fmt.Errorf("Failed to list second level dirs for path %v: %v", path, err)
+						logrus.WithError(err).Warnf("Failed to list volume names for path %v", path)
```
innobead, line 212: the same as the previous, the log could be redundant.

innobead (changes requested):
LGTM, just one comment needs to resolve.

config.go
derekbit: Yes. I think it is fine here. A returned error will show up in longhorn-manager's log. It would be nice to keep a warning message here to give instance-manager logs more context. WDYT?

config.go
derekbit: https://github.com/longhorn/backupstore/pull/138#discussion_r1234680132

config.go
innobead: If it will not cause duplicated logs, I am fine with that.

innobead (approved):

PR #140, opened 2023-06-24
[BACKPORT][v1.4.x] Mark backup job as error if a backup died in the pre-backup stage

derekbit:
Longhorn/longhorn#6182


derekbit:
Not valid

PR #142, opened 2023-06-27
Improve backup log messages

derekbit:


lock.go
```diff
@@ -133,7 +133,7 @@ func (lock *FileLock) Lock() error {
 				if lock.Acquired {
 					if err := saveLock(lock); err != nil {
 						// nothing we can do here, that's why the lock acquisition time is 2x lock refresh interval
-						log.Infof("Failed to refresh acquired lock %v type %v", file, lock.Type)
+						log.WithError(err).Infof("Failed to refresh acquired lock %v type %v", file, lock.Type)
```
innobead, line 136: Need to change log level? Warm?

innobead (changes requested):

lock.go
derekbit: Warn might be better.

innobead (approved):

PR #143, opened 2023-08-11
fix(backup): Relax S3 client retry intervals

mantissahz:
Relax S3 client retry intervals for requests.
Use the default retryer to increase `NumMaxRetries`, `MinRetryDelay`, `MinThrottleDelay`, `MaxRetryDelay` and `MaxThrottleDelay`.

Ref: longhorn/longhorn#2810

### DIFF
--- a/s3/s3_service.go
+++ b/s3/s3_service.go
@@ -5,9 +5,11 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/pkg/errors"
@@ -124,6 +126,14 @@ func (s *Service) PutObject(key string, reader io.ReadSeeker) error {
 		return err
 	}
 	defer s.Close()
+
+	svc.Client.Config.Retryer = client.DefaultRetryer{
+		NumMaxRetries:    10,
+		MinRetryDelay:    500 * time.Millisecond,
+		MinThrottleDelay: 1 * time.Second,
+		MaxRetryDelay:    300 * time.Second,
+		MaxThrottleDelay: 600 * time.Second,
+	}
 
 	params := &s3.PutObjectInput{
 		Bucket: aws.String(s.Bucket),


shuo-wu (commented):
I am thinking if the modified config is enough for a cluster containing hundreds of volumes. Would you have or ask QAs to start a scalability test?

mantissahz:
>I am thinking if the modified config is enough for a cluster containing hundreds of volumes. Would you have or ask QAs to start a scalability test?

Not yet but I did a local test for 200 volumes. cc @longhorn/qa

s3/s3_service.go
```diff
@@ -26,7 +28,16 @@ const (
...
+	config := &aws.Config{
+		Region: &s.Region,
+		Retryer: client.DefaultRetryer{
+			NumMaxRetries:    10,
+			MinRetryDelay:    500 * time.Millisecond,
+			MinThrottleDelay: 1 * time.Second,
+			MaxRetryDelay:    300 * time.Second,
+			MaxThrottleDelay: 600 * time.Second,
+		},
+	}
```
innobead, line 40: This change will impact all operations not just backup, which should be put. 

If what we want to improve is only backup, then let's tune that for put only instead of changing the default setting except we really know all impacts would be. 

You can make the New constructor to have an optional retry parameter, so it can be used if you need a specific retry; otherwise, the current implementation actually already uses the default one which could be generally fine.

BTW, are you able to reproduce the error originally mentioned by Joshua? Is it really a valid issue?

innobead (changes requested):

innobead:
ref: this is the default setting in the s3 lib.

[/vendor/github.com/aws/aws-sdk-go/aws/client/default_retryer.go#L39-L52](https://github.com/longhorn/backupstore/blob/5b4f2d040e9d195d7d27aefc92174f819cb38819/vendor/github.com/aws/aws-sdk-go/aws/client/default_retryer.go#L39-L52)
```go
	// DefaultRetryerMaxNumRetries sets maximum number of retries
	DefaultRetryerMaxNumRetries = 3

	// DefaultRetryerMinRetryDelay sets minimum retry delay
	DefaultRetryerMinRetryDelay = 30 * time.Millisecond

	// DefaultRetryerMinThrottleDelay sets minimum delay when throttled
	DefaultRetryerMinThrottleDelay = 500 * time.Millisecond

	// DefaultRetryerMaxRetryDelay sets maximum retry delay
	DefaultRetryerMaxRetryDelay = 300 * time.Second

	// DefaultRetryerMaxThrottleDelay sets maximum delay when throttled
	DefaultRetryerMaxThrottleDelay = 300 * time.Second
```

innobead:
also, it's interesting to know how the retry has been implemented in the lib as well.

[/vendor/github.com/aws/aws-sdk-go/aws/client/default_retryer.go#L78-L123](https://github.com/longhorn/backupstore/blob/5b4f2d040e9d195d7d27aefc92174f819cb38819/vendor/github.com/aws/aws-sdk-go/aws/client/default_retryer.go#L78-L123)
```go
func (d DefaultRetryer) RetryRules(r *request.Request) time.Duration {

	// if number of max retries is zero, no retries will be performed.
	if d.NumMaxRetries == 0 {
		return 0
	}

	// Sets default value for retryer members
	d.setRetryerDefaults()

	// minDelay is the minimum retryer delay
	minDelay := d.MinRetryDelay

	var initialDelay time.Duration

	isThrottle := r.IsErrorThrottle()
	if isThrottle {
		if delay, ok := getRetryAfterDelay(r); ok {
			initialDelay = delay
		}
		minDelay = d.MinThrottleDelay
	}

	retryCount := r.RetryCount

	// maxDelay the maximum retryer delay
	maxDelay := d.MaxRetryDelay

	if isThrottle {
		maxDelay = d.MaxThrottleDelay
	}

	var delay time.Duration

	// Logic to cap the retry count based on the minDelay provided
	actualRetryCount := int(math.Log2(float64(minDelay))) + 1
	if actualRetryCount < 63-retryCount {
		delay = time.Duration(1<<uint64(retryCount)) * getJitterDelay(minDelay)
		if delay > maxDelay {
			delay = getJitterDelay(maxDelay / 2)
		}
	} else {
		delay = getJitterDelay(maxDelay / 2)
	}
	return delay + initialDelay
}
```

innobead:
@mantissahz any update?

mantissahz:
>@mantissahz any update?

Today will try to change only for the `put` operation instead

innobead:
@mantissahz is this ready for review?

s3/s3_service.go
mantissahz: Modified to only in `put` operations.

>BTW, are you able to reproduce the error originally mentioned by Joshua? Is it really a valid issue?

I can not reproduce it.

mantissahz:
> is this ready for review?
@innobead Yes, it is ready for review. Only modify the retryer in `put` operations.

innobead (approved):

mantissahz:
@Mergifyio backport v1.4.x v1.5.x

mantissahz:
@Mergifyio backport v1.4.x

PR #144, opened 2023-08-13
Add backup common functions and types

derekbit:
The common functions and types were originally in longhorn-engine.

https://github.com/longhorn/longhorn/issues/6514


ChanYiLin (approved):
LGTM

list.go
```diff
@@ -27,7 +27,7 @@ type VolumeInfo struct {
 
 	BackingImageName     string
 	BackingImageChecksum string
-	StorageClassname     string
```
shuo-wu, line 29: Will this affect the marshaled json output and cause a compatibility issue?

derekbit: Oh... you're right.

derekbit: Updated. Thank you.

shuo-wu (approved):

innobead (approved):

innobead:
@derekbit see if the dependant component repos need to do backport too or not. This change should be used by the master, 1.5 and 1.4 of them.

PR #145, opened 2023-08-28
Abstract the disk/lvol file operations in restore functions

derekbit:
https://github.com/longhorn/longhorn/issues/6576


deltablock.go
```diff
@@ -736,8 +729,8 @@ func RestoreDeltaBlockBackup(config *DeltaRestoreConfig) error {
 		// closed.
 		// https://github.com/longhorn/longhorn/issues/2503
 		// We want to truncate regular files, but not device
-		if stat.Mode()&os.ModeType == 0 {
-			log.Debugf("Truncate %v to size %v", volDevName, vol.Size)
+		if stat.Mode().IsRegular() && stat.Mode()&os.ModeType == 0 {
```
innobead, line 735: We can just remove the duplicated `stat.Mode()&os.ModeType == 0` from this if statement. 

```go
func (m FileMode) IsRegular() bool {
	return m&ModeType == 0
}
```

innobead (changes requested):
LGTM, but some comments need to resolve. 

The dependant component repos need to be updated and backported.

deltablock.go
derekbit: Updated. Thank you.

types.go
```diff
@@ -0,0 +1,17 @@
...
+	BLOCKS_DIRECTORY      = "blocks"
+	BLOCK_SEPARATE_LAYER1 = 2
+	BLOCK_SEPARATE_LAYER2 = 4
+	BLK_SUFFIX            = ".blk"
+
+	PROGRESS_PERCENTAGE_BACKUP_SNAPSHOT = 95
+	PROGRESS_PERCENTAGE_BACKUP_TOTAL    = 100
+
+	BackendStoreDriverV1 = "v1"
+	BackendStoreDriverV2 = "v2"
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

derekbit:
cc @innobead

deltablock.go
```diff
@@ -362,7 +373,7 @@ func backupBlock(bsDriver BackupStoreDriver, config *DeltaBackupConfig,
 		return nil
 	}
 
-	log.Debugf("Creating new block file at %v", blkFile)
+	log.Tracef("Creating new block file at %v", blkFile)
```
innobead, line 376: If we can apply the log level setting of longhorn-manager to indirectly apply to all components, then debug level should be enough.

deltablock.go
```diff
@@ -1069,7 +1050,7 @@ func performIncrementalRestore(bsDriver BackupStoreDriver, config *DeltaRestoreC
 
 	errorChans := []<-chan error{errChan}
 	for i := 0; i < int(concurrentLimit); i++ {
-		errorChans = append(errorChans, restoreBlocks(ctx, bsDriver, config, srcVolumeName, blockChan, progress))
+		errorChans = append(errorChans, restoreBlocks(ctx, bsDriver, config.DeltaOps, config.Filename, srcVolumeName, blockChan, progress))
```
innobead, line 1039: So incremental restore will be supported later? or we need to plan for 1.7 instead?

deltablock.go
```diff
@@ -1018,32 +976,41 @@ func restoreBlock(bsDriver BackupStoreDriver, config *DeltaRestoreConfig,
...
+		}()
 
 		for {
 			select {
 			case <-ctx.Done():
+				logrus.Infof("Closing goroutine for restoring blocks for volume %v", volumeName)
+				return
+			case <-deltaOps.GetStopChan():
+				logrus.Infof("Closing goroutine for restoring blocks for %v since received stop signal", volumeName)
+				err = fmt.Errorf("restoration is cancelled since received stop signal")
```
innobead, line 1005: nit: Suggest adding volume name in the error message to ensure the complete error context, even though the caller could enrich the context (volume name) already at the upper layer.

innobead (approved):
LGTM. Just a few comments, but it's good to merge.

deltablock.go
innobead: ref: https://github.com/longhorn/longhorn/issues/6702

PR #146, opened 2023-08-28
feat(backing): backingimage backup support

ChanYiLin:
ref: [longhorn/longhorn#4165](https://github.com/longhorn/longhorn/issues/4165)

For BackingImage backup, we implemented 3 main functions
1. **CreateBackingImageBackup()**
    - with BackupBackingImageStatus interface, we can read the BackingImage data and upload it as block.
2. **RestoreBackingImageBackup()**
    - with RestoreBackingImageStatus interface, we can update the sync_file progress in `backing-image-data-source`
3. **RemoveBackingImageBackup()**

and 3 commands
- **ls-backing-image**: list all backing image names in the backupstore, so we can create the CR with `backup_target_controller`
- **inspect-backing-image**: get the info of the backup backing image, so we can sync the info the CR in the new cluster
- **rm-backing-image**: we can't implement this as grpc in `backing-image-manager`, since there might have no `backing-image-manager` pod in the cluster, so deletion can be only implemented as command and be used in the controller.

### DIFF
--- a/backupbackingimage/backupbackingimage.go
+++ b/backupbackingimage/backupbackingimage.go
@@ -1,0 +1,608 @@
+package backupbackingimage
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"sync"
+
+	"github.com/longhorn/backupstore"
+	"github.com/longhorn/backupstore/common"
+	"github.com/longhorn/backupstore/util"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type BackupBackingImage struct {
+	sync.Mutex
+
+	Name              string
+	Size              int64 `json:",string"`
+	BlockCount        int64 `json:",string"`
+	Checksum          string
+	Labels            map[string]string
+	CompressionMethod string
+	CreatedTime       string
+	CompleteTime      string
+
+	ProcessingBlocks *common.ProcessingBlocks
+
+	Blocks []common.BlockMapping `json:",omitempty"`
+}
+
+type BackupConfig struct {
+	Name            string
+	DestURL         string
+	ConcurrentLimit int32
+}
+
+type RestoreConfig struct {
+	BackupURL       string
+	Filename        string
+	ConcurrentLimit int32
+}
+
+type BackupStatus interface {
+	ReadFile(start int64, data []byte) error
+	CloseFile() error
+	Update(state string, progress int, backupURL string, err string) error
+}
+
+type RestoreStatus interface {
+	UpdateRestoreProgress(progress int, err error)
+}
+
+func CreateBackingImageBackup(config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus, mappings *common.Mappings) (err error) {
+	log := backupstore.GetLog()
+	if config == nil || backupStatus == nil {
+		return fmt.Errorf("invalid empty config or backupStatus for backup")
+	}
+
+	defer func() {
+		if err != nil {
+			backupStatus.Update(string(common.ProgressStateError), 0, "", err.Error())
+		}
+	}()
+
+	bsDriver, err := backupstore.GetBackupStoreDriver(config.DestURL)
+	if err != nil {
+		return err
+	}
+
+	lock, err := backupstore.New(bsDriver, config.Name, backupstore.BACKUP_LOCK)
+	if err != nil {
+		return err
+	}
+
+	defer lock.Unlock()
+	if err := lock.Lock(); err != nil {
+		return err
+	}
+
+	if err := addBackingImage(bsDriver, backupBackingImage); err != nil {
+		return err
+	}
+
+	backupBackingImage, err = loadBackingImage(bsDriver, backupBackingImage.Name)
+	if err != nil {
+		return err
+	}
+
+	log.Info("Creating backup backing image")
+
+	backupBackingImage.Blocks = []common.BlockMapping{}
+	backupBackingImage.ProcessingBlocks = &common.ProcessingBlocks{
+		Blocks: map[string][]*common.BlockMapping{},
+	}
+
+	if err := lock.Lock(); err != nil {
+		return err
+	}
+
+	go func() {
+		defer backupStatus.CloseFile()
+		defer lock.Unlock()
+
+		backupStatus.Update(string(common.ProgressStateInProgress), 0, "", "")
+
+		if progress, backupURL, err := performBackup(bsDriver, config, backupBackingImage, backupStatus, mappings); err != nil {
+			log.WithError(err).Errorf("Failed to perform backup for backing image %v", backupBackingImage.Name)
+			backupStatus.Update(string(common.ProgressStateInProgress), progress, "", err.Error())
+		} else {
+			backupStatus.Update(string(common.ProgressStateInProgress), progress, backupURL, "")
+		}
+	}()
+
+	return nil
+}
+
+func performBackup(bsDriver backupstore.BackupStoreDriver, config *BackupConfig,
+	backupBackingImage *BackupBackingImage, backupStatus BackupStatus, mappings *common.Mappings) (int, string, error) {
+	log := backupstore.GetLog()
+	destURL := config.DestURL
+	concurrentLimit := config.ConcurrentLimit
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	totalBlockCounts, err := getTotalBackupBlockCounts(mappings)
+	if err != nil {
+		return 0, "", err
+	}
+	log.Infof("Creating backup backing image consisting of %v mappings and %v blocks", len(mappings.Mappings), totalBlockCounts)
+
+	progress := &common.Progress{
+		TotalBlockCounts: totalBlockCounts,
+	}
+
+	mappingChan, errChan := common.PopulateMappings(bsDriver, mappings)
+
+	errorChans := []<-chan error{errChan}
+	for i := 0; i < int(concurrentLimit); i++ {
+		errorChans = append(errorChans, backupMappings(ctx, bsDriver, config, backupBackingImage, backupStatus, mappings.BlockSize, progress, mappingChan))
+	}
+	mergedErrChan := common.MergeErrorChannels(ctx, errorChans...)
+	err = <-mergedErrChan
+	if err != nil {
+		return progress.Progress, "", errors.Wrapf(err, "failed to backup backing image %v", backupBackingImage.Name)
+	}
+
+	backupBackingImage.Blocks = common.SortBackupBlocks(backupBackingImage.Blocks, backupBackingImage.Size, mappings.BlockSize)
+	backupBackingImage.CompleteTime = util.Now()
+	backupBackingImage.BlockCount = totalBlockCounts
+	if err := saveBackingImage(bsDriver, backupBackingImage); err != nil {
+		return progress.Progress, "", err
+	}
+
+	return common.ProgressPercentageBackupTotal, EncodeBackupBackingImageURL(config.Name, destURL), nil
+}
+
+func backupMappings(ctx context.Context, bsDriver backupstore.BackupStoreDriver,
+	config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus,
+	blockSize int64, progress *common.Progress, in <-chan common.Mapping) <-chan error {
+
+	errChan := make(chan error, 1)
+	go func() {
+		defer close(errChan)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case mapping, open := <-in:
+				if !open {
+					return
+				}
+
+				if err := backupMapping(bsDriver, config, backupBackingImage, backupStatus, blockSize, mapping, progress); err != nil {
+					errChan <- err
+					return
+				}
+			}
+		}
+	}()
+
+	return errChan
+}
+
+func backupMapping(bsDriver backupstore.BackupStoreDriver,
+	config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus,
+	blockSize int64, mapping common.Mapping, progress *common.Progress) error {
+
+	log := backupstore.GetLog()
+	block := make([]byte, mapping.Size)
+
+	if err := backupStatus.ReadFile(mapping.Offset, block); err != nil {
+		log.WithError(err).Errorf("Failed to read backing image %v block at offset %v size %v", backupBackingImage.Name, mapping.Offset, len(block))
+		return err
+	}
+
+	if err := backupBlock(bsDriver, config, backupBackingImage, backupStatus, mapping.Offset, block, progress); err != nil {
+		logrus.WithError(err).Errorf("Failed to back up backing image %v block at offset %v size %v", backupBackingImage.Name, mapping.Offset, len(block))
+		return err
+	}
+
+	return nil
+}
+
+func backupBlock(bsDriver backupstore.BackupStoreDriver,
+	config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus,
+	offset int64, block []byte, progress *common.Progress) error {
+
+	var err error
+	newBlock := false
+
+	checksum := util.GetChecksum(block)
+
+	if isBlockBeingProcessed(backupBackingImage, offset, checksum) {
+		return nil
+	}
+
+	defer func() {
+		if err != nil {
+			return
+		}
+		backupBackingImage.Lock()
+		defer backupBackingImage.Unlock()
+		updateBlocksAndProgress(backupBackingImage, progress, checksum, newBlock)
+		backupStatus.Update(string(common.ProgressStateInProgress), progress.Progress, "", "")
+	}()
+
+	// skip if block already exists
+	blkFile := getBackingImageBlockFilePath(checksum)
+	if bsDriver.FileExists(blkFile) {
+		return nil
+	}
+
+	newBlock = true
+	rs, err := util.CompressData(backupBackingImage.CompressionMethod, block)
+	if err != nil {
+		return err
+	}
+
+	return bsDriver.Write(blkFile, rs)
+}
+
+func isBlockBeingProcessed(backupBackingImage *BackupBackingImage, offset int64, checksum string) bool {
+	processingBlocks := backupBackingImage.ProcessingBlocks
+
+	processingBlocks.Lock()
+	defer processingBlocks.Unlock()
+
+	blockInfo := &common.BlockMapping{
+		Offset:        offset,
+		BlockChecksum: checksum,
+	}
+	if _, ok := processingBlocks.Blocks[checksum]; ok {
+		processingBlocks.Blocks[checksum] = append(processingBlocks.Blocks[checksum], blockInfo)
+		return true
+	}
+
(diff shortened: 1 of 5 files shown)


ChanYiLin:
Hi @innobead 
Refactored, thanks!

backupbackingimage/config.go
```diff
@@ -0,0 +1,207 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"path/filepath"
+
+	"github.com/longhorn/backupstore"
+	"github.com/longhorn/backupstore/common"
+	"github.com/longhorn/backupstore/util"
+	"github.com/pkg/errors"
```
derekbit, line 11: Reorganize the imports.

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	"fmt"
+	"io"
+	"os"
+	"sync"
+
+	"github.com/longhorn/backupstore"
+	"github.com/longhorn/backupstore/common"
+	"github.com/longhorn/backupstore/util"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
derekbit, line 14: Reorganize imports

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+}
+
+func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile *os.File, decompression string, blk common.BlockMapping) error {
+	blkFile := getBackingImageBlockFilePath(blk.BlockChecksum)
+	rc, err := bsDriver.Read(blkFile)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+	r, err := util.DecompressAndVerify(decompression, rc, blk.BlockChecksum)
```
derekbit, line 471: Off the topic, we can create a ticket for ignoring the error when running into an error.
When one of blocks is somehow corrupted, we can still continue restoring the volume and don't need to error out. Then, users can still rescue the data.

backupbackingimage/backupbackingimage.go
ChanYiLin: fixed, thanks!

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+
+type RestoreStatus interface {
+	UpdateRestoreProgress(progress int, err error)
+}
+
+func CreateBackingImageBackup(config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus, mappings *common.Mappings) (err error) {
+	log := backupstore.GetLog()
+	if config == nil || backupStatus == nil {
+		return fmt.Errorf("invalid empty config or backupStatus for backup")
+	}
```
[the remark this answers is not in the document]
ChanYiLin: fixed!

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+
+	defer func() {
+		if err != nil {
+			backupStatus.Update(string(common.ProgressStateError), 0, "", err.Error())
```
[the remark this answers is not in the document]
ChanYiLin: sure, added.

backupbackingimage/backupbackingimage.go
ChanYiLin: I remember this idea is from this issue: https://github.com/longhorn/longhorn/issues/7069 right?
I think we can reuse that issue to prevent duplication

backupbackingimage/config.go
ChanYiLin: fixed, thanks

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+					return
+				}
+			}
+		}
+	}()
+
+	return errChan
+}
+
+func backupMapping(bsDriver backupstore.BackupStoreDriver,
```
shuo-wu, line 190: NIT: This function is unnecessary. It can be merged with `backupBlock`

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+	if err := lock.Lock(); err != nil {
+		return err
+	}
+	defer lock.Unlock()
+
+	// If we fail to load the backup we still want to proceed with the deletion of the backup file
+	backupBackingImage, err := loadBackingImage(bsDriver, backingImageName)
+	if err != nil {
+		log.WithError(err).Warn("Failed to load to be deleted backup backing image")
```
shuo-wu, line 509: NIT: `..., will continue the deletion"`

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	// check if there have been new backups created while we where processing
+	prevBackupBackingImageNames := backupbackingImageNames
+	backupBackingImageNames, err := GetAllBackupBackingImageNames(bsDriver)
+	if err != nil || !util.UnorderedEqual(prevBackupBackingImageNames, backupBackingImageNames) {
+		log.Info("Found new backup backing image, skip block deletion")
+		return nil
+	}
+
+	// only delete the blocks if it is safe to do so
+	if err := cleanupBlocks(bsDriver, blockInfos); err != nil {
```
shuo-wu, line 546: I am thinking if it's really safe to do this without a lock, even if the backup deletion applies the same flow. What if the backup creation happens during this cleanup?

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	ReadFile(start int64, data []byte) error
+	CloseFile() error
+	Update(state string, progress int, backupURL string, err string) error
+}
+
+type RestoreStatus interface {
+	UpdateRestoreProgress(progress int, err error)
+}
+
+func CreateBackingImageBackup(config *BackupConfig, backupBackingImage *BackupBackingImage, backupStatus BackupStatus, mappings *common.Mappings) (err error) {
```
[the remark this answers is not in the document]
ChanYiLin: Yeah ... I originally thought we can merge BackupVolume and BackupBackingImage together so I move the mapping creation outside the function so this function don't need to know how to read the Volume/BackingImage
 
But it seems quite impossible to merge them together.
Let me move the mapping creation back to this function.

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	defer lock.Unlock()
+	if err := lock.Lock(); err != nil {
+		return err
+	}
+
+	if err := addBackingImage(bsDriver, backupBackingImage); err != nil {
+		return err
+	}
+
+	backupBackingImage, err = loadBackingImage(bsDriver, backupBackingImage.Name)
```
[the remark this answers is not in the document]
ChanYiLin: Fixed, thanks

backupbackingimage/backupbackingimage.go
ChanYiLin: Fixed, thanks!

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
ChanYiLin: You are right, since BackupBackingImage is not like Volume/Backup.
Every BackupBackingImage has its own and only one config. 
If there is config in the store already, we don't need to perform backup for this BackingImage

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+
+	log.Info("Creating backup backing image")
+
+	backupBackingImage.Blocks = []common.BlockMapping{}
+	backupBackingImage.ProcessingBlocks = &common.ProcessingBlocks{
+		Blocks: map[string][]*common.BlockMapping{},
+	}
```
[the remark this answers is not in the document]
ChanYiLin: fixed, thanks

backupbackingimage/backupbackingimage.go
ChanYiLin: merged, thanks

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,608 @@
...
+
+	backupBackingImage, err := loadBackingImage(bsDriver, backingImageName)
+	if err != nil {
+		return errors.Wrapf(err, "backing image %v doesn't exist in backup store", backingImageName)
+	}
+
+	if backupBackingImage.Size == 0 {
+		return fmt.Errorf("read invalid backing image size %v", backupBackingImage.Size)
+	}
```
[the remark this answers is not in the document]
ChanYiLin: Yes, we can check its `CompleteTime` to see if it is completed or not.
If it fails to create backup from the last time, `CompleteTime` will also be empty.

backupbackingimage/backupbackingimage.go
ChanYiLin: We do have lock for `Create`, `Restore`, `Delete`
For `Delete` it is [here](https://github.com/longhorn/backupstore/pull/146/files#diff-ab9eae562809575c7498e159222d9cba875398859f104675b74cc04e23f7ef46R491)

as the lock design mentioned [here](https://github.com/longhorn/longhorn/blob/master/enhancements/20200701-backupstore-file-locks.md)
```
prevent backup deletion during backup restoration
prevent backup deletion while a backup is in progress
prevent backup creation during backup deletion
prevent backup restoration during backup deletion
```

backupbackingimage/backupbackingimage.go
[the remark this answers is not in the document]
ChanYiLin: From my perspective
`backupstore` is a lib imported in `backing-image-manager` to create backingimage backup.
Mapping requires backingimage to preload() to get all the data blocks offset and its size. Currently I put the function in `backing-image-manager` [here](https://github.com/longhorn/backing-image-manager/pull/131/files#diff-90c098feebd583b6d97645040d46c9862f16fffba62b36da5010bd2cfb3e035dR235)

The reason is, from BackupVolume experience, since we put mappings creation in `backupstore`, we then need to make the interface have the functions `OpenSnapshot`, `CompareSnapshot` to create the mapping in the `backupstore` lib and avoid cycle import. And I think that makes the create backup function in `backupstore` not that general.

So I put the mappings creation in `backing-image-manager` and make the interface more general.
In the future, we can even allow other objects to be backed up by the same function simply passing its own mappings and define how to get the block data.

And since functions like `preload()` is implemented in `backing-image-manager`.
Put `CreateBackupBackingImageMappings()` in `backing-image-manager` makes more sense to me.

WDYT?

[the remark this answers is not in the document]
shuo-wu: > In the future, we can even allow other objects to be backed up by the same function by simply passing its own mappings and define how to get the block data.

Good idea. I am fine with the current implementation. And I am thinking if we can modify the volume backup creation function and eliminate the interface you mentioned instead.

shuo-wu (approved):
LGTM

backupbackingimage/backupbackingimage.go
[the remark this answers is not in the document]
ChanYiLin: Yes, I am thinking that as well, but the scope will too big.
We can do the refactor after

backupbackingimage/backupbackingimage.go
innobead: Let's have a ticket for that, but can be referred to 7069 if that is related.

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+	if err != nil {
+		return err
+	}
+	backingImageName, _, err := DecodeBackupBackingImageURL(backupURL)
+	if err != nil {
+		return err
+	}
+	log := backupstore.GetLog()
+	log = log.WithFields(logrus.Fields{"BackingImage": backingImageName})
+
```
innobead, line 490: As I mentioned in another comment, should add this to other functions as if they are specific for a backing image.

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+	}
+	if err := lock.Lock(); err != nil {
+		return err
+	}
+	defer lock.Unlock()
+
+	// If we fail to load the backup we still want to proceed with the deletion of the backup file
+	backupBackingImage, err := loadBackingImageConfigInBackupStore(bsDriver, backingImageName)
+	if err != nil {
+		log.WithError(err).Warn("Failed to load to be deleted backup backing image config, will continue the deletion")
```
innobead, line 503: `"Failed to load the backup backing image config, will continue the deletion")`

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+		return err
+	}
+	log.Info("Removed backup backing image config")
+
+	blockInfos, err := getBlockInfos(bsDriver)
+	if err != nil {
+		return err
+	}
+
+	backupbackingImageNames, err := GetAllBackupBackingImageNames(bsDriver)
```
innobead, line 520: `backupBackingImageNames`

common/backup.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}()
+
+	return blockChan, errChan
+}
+
+// MergeErrorChannels will merge all error channels into a single error out channel.
+// the error out channel will be closed once the ctx is done or all error channels are closed
+// if there is an error on one of the incoming channels the error will be relayed.
+func MergeErrorChannels(ctx context.Context, channels ...<-chan error) <-chan error {
```
innobead, line 116: Very interesting function.

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+	backupBackingImageNames, err := GetAllBackupBackingImageNames(bsDriver)
+	if err != nil || !util.UnorderedEqual(prevBackupBackingImageNames, backupBackingImageNames) {
+		log.Info("Found new backup backing image, skip block deletion")
+		return nil
+	}
+
+	// only delete the blocks if it is safe to do so
+	if err := cleanupBlocks(bsDriver, blockInfos); err != nil {
+		return err
+	}
```
innobead, line 542: If the deletion happens together with another upcoming new backup after `checkAndUpdateBlockInfos` check, how to prevent unexpected block deletion if those blocks are being referenced by a new backup?

How is the current backup deletion?

innobead (changes requested):
In general, LGTM. Some comments need to resolve first.

backupbackingimage/backupbackingimage.go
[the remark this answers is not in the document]
ChanYiLin: Ticket: https://github.com/longhorn/longhorn/issues/7457

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+		BlockChecksum: checksum,
+	}
+	if _, ok := processingBlocks.Blocks[checksum]; ok {
+		processingBlocks.Blocks[checksum] = append(processingBlocks.Blocks[checksum], blockInfo)
+		return true
+	}
+
+	processingBlocks.Blocks[checksum] = []*common.BlockMapping{blockInfo}
+	return false
+}
```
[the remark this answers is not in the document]
ChanYiLin: Yes, this is actually a common function also used in backup volume
I will refactor this part in https://github.com/longhorn/longhorn/issues/7457

common/backup.go
ChanYiLin: This is also a function exists in backup volume https://github.com/longhorn/backupstore/blob/master/util.go#L25
May think about how to refine it in the refactor ticket: https://github.com/longhorn/longhorn/issues/7457

backupbackingimage/backupbackingimage.go
ChanYiLin: There is a lock to prevent it from happening 
https://github.com/longhorn/longhorn/blob/master/enhancements/20200701-backupstore-file-locks.md#proposal

backupbackingimage/backupbackingimage.go
```diff
@@ -0,0 +1,604 @@
...
+type RestoreConfig struct {
+	BackupURL       string
+	Filename        string
+	ConcurrentLimit int32
+}
+
+type BackupStatus interface {
+	ReadFile(start int64, data []byte) error
+	CloseFile() error
+	Update(state string, progress int, backupURL string, err string) error
```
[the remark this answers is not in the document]
ChanYiLin: rename to BackupOperation/RestoreOperation like backup volume

backupbackingimage/backupbackingimage.go
ChanYiLin: ticket: https://github.com/longhorn/longhorn/issues/7458

backupbackingimage/config.go
```diff
@@ -0,0 +1,218 @@
...
+	if !util.ValidateName(backupBackingImage.Name) {
+		return false, fmt.Errorf("invalid backing image name %v", backupBackingImage.Name)
+	}
+
+	if err := saveBackingImageConfig(driver, backupBackingImage); err != nil {
+		return false, errors.Wrapf(err, "failed to add backing image %v", backupBackingImage.Name)
+	}
+
+	log.Infof("Added backupstore backing image %v", backupBackingImage.Name)
+	return false, nil
```
innobead, line 41: Suggest making the log clear, like`log.Infof("Added backupstore backing image config %v", backupBackingImage.Name)` instead of `backing image`. 

Please correct other logs accordingly.

innobead (changes requested):
LGTM. Just one minor log change request.

backupbackingimage/config.go
ChanYiLin: Sure
btw, I think with enriching log fields, we can remove the need of adding the backing image name in the log string

ChanYiLin: Fixed, thanks!

innobead (dismissed):

innobead:
@ChanYiLin Check the failed PR check.

innobead:
The first commit history is weird, and that could be related to how you did the rebase as discussed. I will scratch this PR instead. Please make sure the commit history and rebasing are correct in later PRs.

```
commit e1a25677b56d3f7002b73152f03595cf06f6a48f
Author: Jack Lin <jack.lin@suse.com>
Date:   Mon Aug 28 17:02:29 2023 +0800

    feat(backing): backingimage backup support
    
    ref: longhorn/longhorn 4165
    
    Signed-off-by: Jack Lin <jack.lin@suse.com>
    
    Add code to parse a query at the end of targetURL for NFS mount options.
    
    Signed-off-by: James Munson <james.munson@suse.com>
    
    Correct construction of backupURLs.
    
    Signed-off-by: James Munson <james.munson@suse.com>
    
    Address review feedback.  Add CIFS code.  Expand unit tests.
    
    Signed-off-by: James Munson <james.munson@suse.com>
    
    Revise to always keep nfsOptions during Encode and Decode BackupURL.
    
    Signed-off-by: James Munson <james.munson@suse.com>
    
    Add output to ensure test is being run.
    
    Signed-off-by: James Munson <james.munson@suse.com>
    
    feat: update golang to v1.21
    
    Signed-off-by: PoAn Yang <poan.yang@suse.com>
    
    fix(backup): Relax S3 client retry intervals
    ....
```

PR #149, opened 2023-10-11
NFS backup target mount options

james-munson:
https://github.com/longhorn/longhorn/issues/6608

Accept NFS mount options as a query appended to the backup target URL.

### DIFF
--- a/backupstore.go
+++ b/backupstore.go
@@ -112,12 +112,18 @@ func removeVolume(volumeName string, driver BackupStoreDriver) error {
 }
 
 func EncodeBackupURL(backupName, volumeName, destURL string) string {
+	u, err := url.Parse(destURL)
+	if err != nil {
+		return ""
+	}
 	v := url.Values{}
 	v.Add("volume", volumeName)
 	if backupName != "" {
 		v.Add("backup", backupName)
 	}
-	return destURL + "?" + v.Encode()
+	// This will drop any Query tag (such as "nfsOptions") and replace with just the backup name tags.
+	u.RawQuery = v.Encode()
+	return u.String()
 }
 
 func DecodeBackupURL(backupURL string) (string, string, string, error) {
@@ -134,6 +140,9 @@ func DecodeBackupURL(backupURL string) (string, string, string, error) {
 	if backupName != "" && !util.ValidateName(backupName) {
 		return "", "", "", fmt.Errorf("invalid backup name parsed, got %v", backupName)
 	}
+	// Is this going to cause a problem because we are discarding any other Query tags here
+	// rather than just removing "volume" and "backup"?
+	// No, because no caller anywhere uses the returned destURL value anyway.
 	u.RawQuery = ""
 	destURL := u.String()
 	return backupName, volumeName, destURL, nil
--- a/backupstore.go
+++ b/backupstore.go
@@ -116,12 +116,18 @@ func EncodeBackupURL(backupName, volumeName, destURL string) string {
 	if err != nil {
 		return ""
 	}
-	v := url.Values{}
+
+	v, err := url.ParseQuery(u.RawQuery)
+	if err != nil {
+		// Just start with empty values list then
+		v = url.Values{}
+	}
+
 	v.Add("volume", volumeName)
 	if backupName != "" {
 		v.Add("backup", backupName)
 	}
-	// This will drop any Query tag (such as "nfsOptions") and replace with just the backup name tags.
+
 	u.RawQuery = v.Encode()
 	return u.String()
 }
@@ -140,10 +146,10 @@ func DecodeBackupURL(backupURL string) (string, string, string, error) {
 	if backupName != "" && !util.ValidateName(backupName) {
 		return "", "", "", fmt.Errorf("invalid backup name parsed, got %v", backupName)
 	}
-	// Is this going to cause a problem because we are discarding any other Query tags here
-	// rather than just removing "volume" and "backup"?
-	// No, because no caller anywhere uses the returned destURL value anyway.
-	u.RawQuery = ""
+
+	v.Del("volume")
+	v.Del("backup")
+	u.RawQuery = v.Encode()
 	destURL := u.String()
 	return backupName, volumeName, destURL, nil
 }
--- a/backupstore_test.go
+++ b/backupstore_test.go
@@ -14,24 +14,27 @@ func TestEncodeAndDecodeBackupURL(t *testing.T) {
 		destURL           string
 		expectDecodeError bool
 		expectBackupURL   string
+		expectDecodeURL   string
 	}{
 		{
 			volumeName:      "vol-1",
 			destURL:         "s3://backupstore@minio/",
 			expectBackupURL: "s3://backupstore@minio/?volume=vol-1",
+			expectDecodeURL: "s3://backupstore@minio/",
 		},
 		{
 			volumeName:      "vol-2",
 			backupName:      "backup-2",
 			destURL:         "s3://backupstore@minio/",
 			expectBackupURL: "s3://backupstore@minio/?backup=backup-2&volume=vol-2",
+			expectDecodeURL: "s3://backupstore@minio/",
 		},
 		{
 			// Test invalid volume name
 			volumeName:        "-3-vol",
 			destURL:           "s3://backupstore@minio/",
-			expectDecodeError: true,
 			expectBackupURL:   "s3://backupstore@minio/?volume=-3-vol",
+			expectDecodeError: true,
 		},
 		{
 			// Test invalid backup name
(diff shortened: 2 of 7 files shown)


backupstore_test.go
```diff
@@ -41,6 +44,22 @@ func TestEncodeAndDecodeBackupURL(t *testing.T) {
...
+			expectDecodeURL: "nfs://longhorn-test-nfs-svc.default:/opt/backupstore",
+		},
+		{
+			// Test NFS target with mount options (dropped by DecodeBackupURL.)
+			volumeName:      "vol-6",
+			backupName:      "backup-6",
+			destURL:         "nfs://longhorn-test-nfs-svc.default:/opt/backupstore?nfsOptions=soft,timeo=150,retrans=3",
+			expectBackupURL: "nfs://longhorn-test-nfs-svc.default:/opt/backupstore?backup=backup-6&volume=vol-6",
+			expectDecodeURL: "nfs://longhorn-test-nfs-svc.default:/opt/backupstore",
+		},
```
PhanLe1010, line 62: Thank you for adding this test! It helps to understand the implementation

backupstore.go
```diff
@@ -132,6 +138,9 @@ func DecodeBackupURL(backupURL string) (string, string, string, error) {
 	if backupName != "" && !util.ValidateName(backupName) {
 		return "", "", "", fmt.Errorf("invalid backup name parsed, got %v", backupName)
 	}
+	// Is this going to cause a problem because we are discarding any other Query tags here
+	// rather than just removing "volume" and "backup"?
+	// No, because no caller anywhere uses the returned destURL value anyway.
```
PhanLe1010, line 143: I am wondering if we should not return the destURL at all because no caller is using it.

nfs/nfs.go
```diff
@@ -72,8 +73,20 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
...
 
+	nfsOptions, exist := u.Query()["nfsOptions"]
+	if exist {
+		if len(nfsOptions) > 1 {
+			// Options in the form "nfsOptions=soft&nfsOptions=timeo=450&nfsOptions=retrans=3" are legal.
+			b.mountOptions = nfsOptions
+		} else {
+			// Options in the form "nfsOptions=soft,timeo=450,retrans=3" are more likely, but we must split them.
+			b.mountOptions = strings.Split(nfsOptions[0], ",")
+		}
```
PhanLe1010, line 84: Can we add a test case with backup url  `nfs://longhorn-test-nfs-svc.default:/opt/backupstore?nfsOptions=` to see if the code here can be panic (out of range panic)  as the `nfsOptions` slice may have 0 length?

nfs/nfs.go
```diff
@@ -72,8 +73,20 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 	b.destURL = KIND + "://" + b.serverPath
 	b.mountDir = filepath.Join(util.MountDir, strings.TrimRight(strings.Replace(u.Host, ".", "_", -1), ":"), u.Path)
 
+	nfsOptions, exist := u.Query()["nfsOptions"]
+	if exist {
+		if len(nfsOptions) > 1 {
+			// Options in the form "nfsOptions=soft&nfsOptions=timeo=450&nfsOptions=retrans=3" are legal.
```
PhanLe1010, line 79: I am wondering if we should force use of one format only to reduce the error surface. I.e., use the format `"nfsOptions=soft,timeo=450,retrans=3"` only

nfs/nfs.go
james-munson: I tested that in a little standalone app.  There is no panic.  What the parser does in that case is make an array of strings of length 1, with the single string "", so len(nfsOptions) is never 0.
With the current logic, when that is detected, the non-empty array of strings will be used in place of the default array of options, and give the net result of using no mount options at all.  Which is a reasonable behavior.

backupstore.go
james-munson: Yes, I think it was there originally just in case.  But I have having second thoughts about discarding the other Query tags.  I think that may lead us to lose the nfsOptions in some situations, and to change how we mount if we happen not to be mounted and start by looking for a backed-up volume.
I think the answer is to always keep the other tags, and ensure that every caller goes through the Encode and Decode API to do the right thing.

nfs/nfs.go
james-munson: I found the other format when I was working around the longhorn-manager's check of legal URLs, which mistakenly rejects ",".  So it will work even before that fix is promoted: https://github.com/longhorn/longhorn-manager/pull/2216.  (Thanks for the review on that!)

I see your point, but in this case, I don't think the exposure is all that great, and it is a legal URL, so I am inclined to leave it.

nfs/nfs.go
```diff
@@ -97,27 +110,44 @@ func (b *BackupStoreDriver) mount() error {
...
-
-		mountOptions := []string{
-			fmt.Sprintf("nfsvers=%v", version),
-			"actimeo=1",
-			"soft",
-			"timeo=300",
-			"retry=2",
-		}
+	// If overridden, assume nfsvers is specified or defaulted, but don't try all minor versions.
+	if len(b.mountOptions) > 0 {
```
shuo-wu, line 114: NIT: Could we detect if option `nfsvers` is already specified by users? We may still need to try all minor versions when the option is not in this list.

nfs/nfs.go
shuo-wu: NIT: You can add cases for both kinds of format in the unit test.

shuo-wu (approved):
In general LGTM
BTW, will we support the mount option for CIFS later?

backupstore.go
ejweber, line 1: I was reviewing the NFS man page, and it seems to imply that the `nfsvers` mount option should not be used with the `nfs4` type. Maybe we should be using `minorversion` or `vers` instead? It reads like it probably works anyway, so this is likely just a NIT.

https://linux.die.net/man/5/nfs

nfs/nfs.go
```diff
@@ -97,27 +110,44 @@ func (b *BackupStoreDriver) mount() error {
...
 
-		err := util.MountWithTimeout(mounter, b.serverPath, b.mountDir, "nfs4", mountOptions, sensitiveMountOptions,
+		err := util.MountWithTimeout(mounter, b.serverPath, b.mountDir, "nfs4", b.mountOptions, sensitiveMountOptions,
 			defaultMountInterval, defaultMountTimeout)
 		if err == nil {
 			return nil
 		}
 
-		retErr = errors.Wrapf(retErr, "vers=%s: %v", version, err.Error())
+		retErr = errors.Wrapf(retErr, "version defaulted: %v", err.Error())
```
ejweber, line 125: NIT: If I understand correctly, we may have used a specific version specified by the user, and not a default version. Should we change the wording here (or return the version we actually used if it was specified)?

nfs/nfs.go
ejweber: I'm not sure about this. I am thinking the logic should be as follows:

If the user specifies ANY mount options, they become fully responsible for specifying ALL mount options. We should not try to play with options on our end to make things work. This makes our behavior more predictable (and hopefully prevents us from escalating our support of custom mount options to validating a wide variety of things the user might input).

ejweber (commented):
In general, LGTM. A couple of minor NITs for discussion.

backupstore.go
PhanLe1010: This is being discussed in our internal slack also:

> I have a concern about keeping the mount options parameters in the backupURL which is recorded in the backup CR:
> 
> * There could be multiple clusters connecting to the same backup target. Can we double check that setting from one cluster doesn't overwrite setting in other cluster?
> * Also, there will be some backup created before user set the nfsOptions parameters. Does it create an inconsistency as accessing new backups will be mounted differently from old backups?
>

nfs/nfs.go
PhanLe1010: Sounds good to me. Thanks for double-checking it

nfs/nfs.go
PhanLe1010: Sounds good to me!

backupstore.go
james-munson: Agreed, but it does appear to work, and that code was already here, so we'll leave it as the "legacy path".

backupstore.go
james-munson: Phan makes good points about this, so I am leaving as I first did it.  In principle, the unused return could be eliminated, but that would also require changes in longhorn-engine and longhorn-manager.

james-munson:
Updated the PR to address review notes, and added the CIFS code (simple enough.)

PhanLe1010:
@james-munson and I discussed in the slack channel about a concern. 

With the current implementation, when the user specifies the nfs options, I am wondering if it is possible that sometime we mount the nfs with the options and sometime not depending on the operation ordering:

**Case 1:**
1. Longhorn manager pod executes this function first https://github.dev/longhorn/longhorn-manager/blob/286e6e600ec9e47fec3be0e527f60f0ecc4cb4af/engineapi/backups.go#L165-L175.
1. It will mount the nfs with the specified options

**Case 2:**
1. Longhorn manager pod executes this function first https://github.dev/longhorn/longhorn-manager/blob/286e6e600ec9e47fec3be0e527f60f0ecc4cb4af/engineapi/backups.go#L258-L269
1. The URL here is the returned value of the EncodeBackupURL function which strips away the nfs options https://github.dev/longhorn/longhorn-manager/blob/286e6e600ec9e47fec3be0e527f60f0ecc4cb4af/controller/backup_controller.go#L415-L416
1. It will mount the nfs **without** any options

james-munson:
To Phan's point - looking through the code, about half the paths that call GetBackupStoreDriver do so with a backup URL that ultimately came from a call to EncodeBackupURL.  If we make sure not to discard the mount options settings at any step, then we should be covered in either case above.  That's the best we can do.  It's true that legacy backupURLs stored elsewhere might contain no options, and if one of those is used for the first mount, then we would use the default options.  I don't think there is much we can do about that.

I have revised the code to always use the full URL when doing Encode and Decode, only adding and removing the tags for volume and backup name. I think that's the best answer.

backupstore.go
james-munson: Superseded by the note below.

PhanLe1010 (approved):
LGTM. 

Thank you!

ejweber (approved):
LGTM.

Thanks for sticking with this one! A little bit of churn, but I think we have a safer PR for it (and I learned quite a bit about our backup functionality in the process).

nfs/nfs.go
```diff
@@ -72,11 +74,18 @@ func initFunc(destURL string) (backupstore.BackupStoreDriver, error) {
 	b.destURL = KIND + "://" + b.serverPath
 	b.mountDir = filepath.Join(util.MountDir, strings.TrimRight(strings.Replace(u.Host, ".", "_", -1), ":"), u.Path)
 
+	nfsOptions, exist := u.Query()["nfsOptions"]
```
derekbit, line 77: nit: using `options` rather than `nfsOptions` is simple and enough. same for `cifsOptions`.

james-munson:
@mergify backport v1.5.x

james-munson:
@innobead This is odd - there is no v1.5.x branch.

innobead:
Our lib should be backward compatible, so there is no 1.5 branch. However, we had an exception for 1.4/1.3, since we didn't ensure backward compatibility at that time.

Thus, there's no need to backport this to 1.5.x branch, but just update the lib dependency of dependent components (longhorn manager, longhorn engine, etc) of master and 1.5 branches.

PR #155, opened 2024-01-09
Add context parameter to RestoreDeltaBlockBackup function

derekbit:

Make the goroutine created in `RestoreDeltaBlockBackup `stoppable by an external context.


derekbit:
Wrong branch

PR #156, opened 2024-01-09
Make restore goroutine stoppable by an external context

derekbit:

Make restore goroutine stoppable by an external context

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -607,7 +607,7 @@ func mergeSnapshotMap(deltaBackup, lastBackup *Backup) *Backup {
 }
 
 // RestoreDeltaBlockBackup restores a delta block backup for the given configuration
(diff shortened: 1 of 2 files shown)


derekbit:
Ready for review, but CodeFactor seems problematic.
image

cc @innobead

innobead (approved):
LGTM. 

For any operations including blocking/sync or async, they should be cancellable.

innobead:
> Ready for review, but CodeFactor seems problematic

Perhaps that issue could be related to the ongoing GitHub instability.
https://www.githubstatus.com/

image

derekbit:
> LGTM.
> 
> For any operations including blocking/sync or async, they should be cancellable.

Yeah, it looks `backupstore` never handles cancellation.
The PR focuses on restore cancelation. I will submit another PR for backup.

PR #157, opened 2024-01-10
Issue 7581

derekbit:

1. Return appropriate error message if backup restoration is canceled.
2. Introduce `ProgressStateUndefined` and `ProgressStateCanceled` states.


deltablock.go
```diff
@@ -996,11 +996,12 @@ func restoreBlocks(ctx context.Context, bsDriver BackupStoreDriver, deltaOps Del
...
 			case <-ctx.Done():
-				logrus.Infof("Closing goroutine for restoring blocks for volume %v", volumeName)
+				logrus.Infof("Closing goroutine for restoring blocks for volume %v since server stop", volumeName)
+				err = fmt.Errorf(types.ErrorMsgRestoreCancelled + " since server stop")
 				return
 			case <-deltaOps.GetStopChan():
-				logrus.Infof("Closing goroutine for restoring blocks for %v since received stop signal", volumeName)
-				err = fmt.Errorf("restoration is cancelled since received stop signal")
+				logrus.Infof("Closing goroutine for restoring blocks for volume %v since received stop signal", volumeName)
+				err = fmt.Errorf(types.ErrorMsgRestoreCancelled + " since received stop signal")
```
innobead, line 1004: The same here, but the error message should be different, as the trigger point is different.

innobead (changes requested):

innobead (approved):

PR #158, opened 2024-01-25
fix: skip remove backup volume if the dir is already gone

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7744

There is a chance that `BackupVolumeDelete()` will be executed twice even the finalizer of the BackupVolume is already gone.
Every time `DeleteBackupVolume()` is executed, it will create a lock file.
It is possible that during the second execution, the nfs is unmount because the BackupTarget is unset after BackupVolumes are all deleted in some automated operations (e.g. longhorn e2e test)
So the lock remains on nfs and it pollutes the env.

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -1051,6 +1051,12 @@ func DeleteBackupVolume(volumeName string, destURL string) error {
 	if err != nil {
 		return err
 	}
+
+	// no need to lock and remove volume if it does not exist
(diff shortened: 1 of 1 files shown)


derekbit (dismissed):
LGTM

ChanYiLin:
Should use bsdriver instead of local `os` to check fileExists

innobead (approved):

PR #160, opened 2024-01-29
fix: skip deletion if folder does not exist

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7744

Since s3 doesn't support checking if folder exists.
We can use existing `getVolumeNames` which "returns all volume names based on the folders on the backupstore"
to see if the folder exists or not.

### DIFF
--- a/deltablock.go
+++ b/deltablock.go
@@ -1051,6 +1051,17 @@ func DeleteBackupVolume(volumeName string, destURL string) error {
 	if err != nil {
 		return err
 	}
+
+	backupVolumeFolderExists, err := volumeFolderExists(bsDriver, volumeName)
+	if err != nil {
+		return err
+	}
+
+	// No need to lock and remove volume if it does not exist.
+	if !backupVolumeFolderExists {
+		return nil
+	}
+
 	lock, err := New(bsDriver, volumeName, DELETION_LOCK)
 	if err != nil {
 		return err
--- a/config.go
+++ b/config.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


ChanYiLin:
## Test

test_basic: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6172/console
test_recurring_job_restored_from_backup_target: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6171/console

deltablock.go
```diff
@@ -1051,6 +1053,30 @@ func DeleteBackupVolume(volumeName string, destURL string) error {
 	if err != nil {
 		return err
 	}
+
+	// Check if volume folder exists on backupstore
```
derekbit, line 1057: Can we move the codes into a function `DoesBackupVolumeFolderExists()` for simplifying the function?

ChanYiLin: @derekbit Thanks, fixed.

innobead:
Let's make this after 1.6.0 if this is not a necessary fix. WDYT?

ChanYiLin:
> Let's make this after 1.6.0 if this is not a necessary fix. WDYT?

@innobead Agree, we can merge it later

innobead (changes requested):
Why not just make `New` lock factory function call fail with an expected error (ex: UnsupportedLockError) and bypass the error if the driver is not filesystem-based? This way, you don't need to do any volume folder checking which needs extra resource consumption. 

[/deltablock.go#L1054-L1063](https://github.com/longhorn/backupstore/blob/9c90a8728ca2e1f95f10a4c6cf67cf7ac20c23e4/deltablock.go#L1054-L1063)
```go
	lock, err := New(bsDriver, volumeName, DELETION_LOCK)
	if err != nil {
		return err
	}

		return err
	}
	defer lock.Unlock()
	return removeVolume(volumeName, bsDriver)
```

innobead (approved):

innobead:
@ChanYiLin already explained the context. My previous review is invalid. LGTM.

PR #166, opened 2024-03-19
feat: support full backup

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7070

- Add `BackupCount` in BackupVolume, so we can calculate the interval to achieve periodically full backup 
- Add `NewlyUploadedDataSize`  and `ReUploadedDataSize` to each Backup. So user can know the data transfer size for each Backup to trace the cost.
- Add `Parameters` to the Backup Config. 
    - If the backup has the parameter `backup-mode: full`, it will pretend there is no last backup, and backup all the blocks of the current volume.


ChanYiLin:
## Dependent PR:
- https://github.com/longhorn/go-common-libs/pull/26

ChanYiLin:
CI failed, will consult @FrankYang0529 
```
Error response from daemon: {"message":"client version 1.22 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version"}
Starting NFS server
Warning: failed to get default registry endpoint from daemon (Error response from daemon: {"message":"client version 1.22 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version"}). Using system default: https://index.docker.io/v1/
Error response from daemon: {"message":"client version 1.22 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version"}
```

backupstore.go
```diff
@@ -37,15 +37,18 @@ type ProcessingBlocks struct {
...
+	SnapshotName          string
+	SnapshotCreatedAt     string
+	CreatedTime           string
+	Size                  int64 `json:",string"`
+	Labels                map[string]string
+	Parameters            map[string]string
+	IsIncremental         bool
+	CompressionMethod     string
+	NewlyUploadedDataSize int64 `json:",string"`
+	ReUploadedDataSize    int64 `json:",string"`
```
derekbit, line 51: Nit: As we discussed before, there are some fields that can be part of parameters. The parameters are like metadata or attributes. We can reorganize the struct later.

ChanYiLin: Hi @derekbit 
https://github.com/longhorn/longhorn/issues/8791

deltablock.go
```diff
@@ -401,21 +403,57 @@ func backupBlock(bsDriver BackupStoreDriver, config *DeltaBackupConfig,
...
 	rs, err := util.CompressData(deltaBackup.CompressionMethod, block)
 	if err != nil {
 		return err
 	}
 
+	dataSize, err := getTransferDataSize(rs)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get transfer data size during saving blocks")
+	}
+	updateUploadDataSize(reUpload, deltaBackup, dataSize)
```
derekbit, line 427: Can we write data first and then update the upload size instead?

derekbit (commented):
In general, LGTM.
Only few comments needing to address.

derekbit (approved):
LGTM

PR #167, opened 2024-04-17
Fall back to other decompression method if the current one fails

derekbit:

Issue Longhorn/longhorn#7687

The mitigation will be removed after identifying https://github.com/longhorn/longhorn/issues/7687

### DIFF
--- a/backupbackingimage/backupbackingimage.go
+++ b/backupbackingimage/backupbackingimage.go
@@ -482,9 +482,11 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
 		return err
 	}
 	defer rc.Close()
-	r, err := util.DecompressAndVerify(decompression, rc, blk.BlockChecksum)
+	r, err := util.DecompressAndVerifyWithFallback(decompression, rc, blk.BlockChecksum)
 	if err != nil {
-		return err
+		if r == nil {
+			return err
+		}
 	}
 
 	if _, err := backingImageFile.Seek(blk.Offset, 0); err != nil {
--- a/deltablock.go
+++ b/deltablock.go
@@ -252,6 +252,10 @@ func CreateDeltaBlockBackup(backupName string, config *DeltaBackupConfig) (isInc
 		},
 	}
 
+	log = logrus.WithFields(logrus.Fields{
+		"compressionMethod": volume.CompressionMethod,
+	})
+
 	// keep lock alive for async go routine.
 	if err := lock.Lock(); err != nil {
 		deltaOps.CloseSnapshot(snapshot.Name, volume.Name)
@@ -264,6 +268,7 @@ func CreateDeltaBlockBackup(backupName string, config *DeltaBackupConfig) (isInc
 		deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), 0, "", "")
 
 		log.Info("Performing delta block backup")
+
 		if progress, backup, err := performBackup(bsDriver, config, delta, deltaBackup, backupRequest.lastBackup); err != nil {
 			logrus.WithError(err).Errorf("Failed to perform backup for volume %v snapshot %v", volume.Name, snapshot.Name)
 			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), progress, "", err.Error())
@@ -738,18 +743,22 @@ func restoreBlockToFile(bsDriver BackupStoreDriver, volumeName string, volDev *o
 	blkFile := getBlockFilePath(volumeName, blk.BlockChecksum)
 	rc, err := bsDriver.Read(blkFile)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to read block %v with checksum %v", blkFile, blk.BlockChecksum)
 	}
 	defer rc.Close()
-	r, err := util.DecompressAndVerify(decompression, rc, blk.BlockChecksum)
+
+	r, err := util.DecompressAndVerifyWithFallback(decompression, rc, blk.BlockChecksum)
 	if err != nil {
-		return err
+		if r == nil {
+			return err
+		}
 	}
+
 	if _, err := volDev.Seek(blk.Offset, 0); err != nil {
-		return err
+		return errors.Wrapf(err, "failed to seek to offset %v for decompressed block %v", blk.Offset, blkFile)
 	}
 	_, err = io.CopyN(volDev, r, DEFAULT_BLOCK_SIZE)
-	return err
+	return errors.Wrapf(err, "failed to write decompressed block %v to volume %v", blkFile, volumeName)
 }
 
 func RestoreDeltaBlockBackupIncrementally(ctx context.Context, config *DeltaRestoreConfig) error {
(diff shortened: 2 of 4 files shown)


backupbackingimage/backupbackingimage.go
```diff
@@ -484,7 +486,26 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
 	defer rc.Close()
 	r, err := util.DecompressAndVerify(decompression, rc, blk.BlockChecksum)
 	if err != nil {
-		return err
+		// Fall back to other decompression method if the current one fails
+		// The mitigation will be removed after identifying https://github.com/longhorn/longhorn/issues/7687
+		// Seek rc to offset 0
+		seeker, ok := rc.(io.Seeker)
+		if !ok {
+			return errors.New("reader does not support seeking")
```
innobead, line 494: Please wrap the original err as well, so we can know the source error which is related to the wrong compression method.

ChanYiLin (dismissed):
LGTM

backupbackingimage/backupbackingimage.go
```diff
@@ -484,7 +486,26 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
...
+		}
+
+		if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+			return errors.Wrapf(err, "failed to seek to offset 0 for block %v", blk.BlockChecksum)
+		}
```
innobead, line 499: Warp the original err. Review other parts in this if clause to ensure the original error will be included.

deltablock.go
```diff
@@ -738,18 +745,39 @@ func restoreBlockToFile(bsDriver BackupStoreDriver, volumeName string, volDev *o
...
+
+		if strings.Contains(err.Error(), gzip.ErrHeader.Error()) {
+			r, err = util.DecompressAndVerify("lz4", rc, blk.BlockChecksum)
+		} else if strings.Contains(err.Error(), "lz4: bad magic number") {
+			r, err = util.DecompressAndVerify("gzip", rc, blk.BlockChecksum)
+		}
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to decompress and verify block %v", blk.BlockChecksum)
 	}
```
c3y1huang, line 774: Suggest moving this in the block where the alternative decompression are attempted.
```suggestion
		}
		if err != nil {
			return errors.Wrapf(err, "failed to decompress and verify block %v", blk.BlockChecksum)
		}
	}
```

backupbackingimage/backupbackingimage.go
```diff
@@ -484,7 +486,26 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
...
+
+		if strings.Contains(err.Error(), gzip.ErrHeader.Error()) {
+			r, err = util.DecompressAndVerify("lz4", rc, blk.BlockChecksum)
+		} else if strings.Contains(err.Error(), "lz4: bad magic number") {
+			r, err = util.DecompressAndVerify("gzip", rc, blk.BlockChecksum)
+		}
+	}
+	if err != nil {
+		return errors.Wrapf(err, "failed to decompress and verify block %v", blk.BlockChecksum)
 	}
```
c3y1huang, line 509: Should we have a function for this to avoid duplicating logic?

backupbackingimage/backupbackingimage.go
```diff
@@ -484,7 +486,26 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
...
+
+		if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+			return errors.Wrapf(err, "failed to seek to offset 0 for block %v", blk.BlockChecksum)
+		}
+
+		if strings.Contains(err.Error(), gzip.ErrHeader.Error()) {
+			r, err = util.DecompressAndVerify("lz4", rc, blk.BlockChecksum)
+		} else if strings.Contains(err.Error(), "lz4: bad magic number") {
+			r, err = util.DecompressAndVerify("gzip", rc, blk.BlockChecksum)
+		}
```
innobead, line 505: it seems we only provide `lz4` and `gzip`, so probably just use `else` instead of `else if` clause here? as `gzip` will be our final default compression method.

c3y1huang (changes requested):
Drone failed, could you check?

innobead (changes requested):
In general, LGTM. 

Please update the suggestion for backing image resotre abdn backup restore parts.

backupbackingimage/backupbackingimage.go
```diff
@@ -484,7 +486,26 @@ func restoreBlockToFile(bsDriver backupstore.BackupStoreDriver, backingImageFile
 	defer rc.Close()
 	r, err := util.DecompressAndVerify(decompression, rc, blk.BlockChecksum)
 	if err != nil {
-		return err
+		// Fall back to other decompression method if the current one fails
```
c3y1huang, line 489: Should we have a function for the alternative decompression to avoid duplicating logic?

util/util.go
```diff
@@ -120,19 +120,48 @@ func CompressData(method string, data []byte) (io.ReadSeeker, error) {
...
-		return nil, err
+		return nil, errors.Wrap(err, "failed to read decompressed data")
 	}
 	if GetChecksum(block) != checksum {
 		return nil, fmt.Errorf("checksum verification failed for block")
 	}
 	return bytes.NewReader(block), nil
 }
 
+func DecompressAndVerifyWithFallback(decompression string, rc io.ReadCloser, checksum string) (io.Reader, error) {
```
innobead, line 136: nit: suggest adding unit tests

innobead (approved):
LGTM

util/util.go
derekbit: All compression and decompression codes need unit tests.
Let's handle this later.
WDYT?

derekbit: We have `io.seek` in the function. Thinking how to address it in a simple way.
I will add the unit test later in another PR.
WDYT?

innobead: Sounds good.

deltablock.go
```diff
@@ -738,18 +743,20 @@ func restoreBlockToFile(bsDriver BackupStoreDriver, volumeName string, volDev *o
 	blkFile := getBlockFilePath(volumeName, blk.BlockChecksum)
 	rc, err := bsDriver.Read(blkFile)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to read block %v from backupstore %v", blk.BlockChecksum, blkFile)
```
innobead, line 746: The message seems a little confusing. 

How about? Please check if other error messages need to adjust. 

```go
errors.Wrapf(err, "failed to read block %v with checksum %v from backupstore", blkFile, blk.BlockChecksum)
```

util/util.go
```diff
@@ -120,19 +120,50 @@ func CompressData(method string, data []byte) (io.ReadSeeker, error) {
...
+	if strings.Contains(err.Error(), gzip.ErrHeader.Error()) {
+		r, errFallback = DecompressAndVerify("lz4", rc, checksum)
+	} else if strings.Contains(err.Error(), "lz4: bad magic number") {
+		r, errFallback = DecompressAndVerify("gzip", rc, checksum)
+	}
+	if errFallback != nil {
+		return nil, errors.Wrapf(multierr.Append(err, errFallback), "failed to decompress and verify block %v with fallback", checksum)
+	}
+
+	return r, nil
```
innobead, line 164: Is it possible that `err` is not nil, but `errFallback` is nil?

Suggest to the below change:
```go
return r, err
```

util/util.go
innobead: nit: Add a function description.

util/util.go
derekbit: Addressed. Thank you.

innobead (approved):

PR #170, opened 2024-04-26
fix: golangci-lint error

FrankYang0529:
https://github.com/longhorn/longhorn/issues/7448


backupbackingimage/backupbackingimage.go
```diff
@@ -298,93 +306,105 @@
...
 		return err
 	}
 
-	defer lock.Unlock()
 	if err := lock.Lock(); err != nil {
 		return err
 	}
+	defer func() {
+		if unlockErr := lock.Unlock(); unlockErr != nil {
+			logrus.WithError(unlockErr).Warn("Failed to unlock restore")
```
innobead, line 338: Suggest log: `Failed to unlock restore backing image` 

Please correct other logs if needed.

deltablock.go
```diff
@@ -258,22 +268,38 @@ func CreateDeltaBlockBackup(backupName string, config *DeltaBackupConfig) (isInc
...
 		if progress, backup, err := performBackup(bsDriver, config, delta, deltaBackup, backupRequest.lastBackup); err != nil {
 			logrus.WithError(err).Errorf("Failed to perform backup for volume %v snapshot %v", volume.Name, snapshot.Name)
-			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), progress, "", err.Error())
+			if updateErr := deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), progress, "", err.Error()); updateErr != nil {
+				logrus.WithError(updateErr).Error("Failed to update backup status")
+			}
 		} else {
-			deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), progress, backup, "")
+			if updateErr := deltaOps.UpdateBackupStatus(snapshot.Name, volume.Name, string(types.ProgressStateInProgress), progress, backup, ""); updateErr != nil {
+				logrus.WithError(updateErr).Error("Failed to update backup status")
```
innobead, line 301: It seems no hurt and not critical if encountering this error. Should we just print it in warn level?

fsops/fsops.go
```diff
@@ -127,7 +127,9 @@ func (f *FileSystemOperator) List(path string) ([]string, error) {
 func (f *FileSystemOperator) Upload(src, dst string) error {
 	tmpDst := dst + ".tmp" + "." + strconv.FormatInt(time.Now().UTC().UnixNano(), 10)
 	if f.FileExists(tmpDst) {
-		f.Remove(tmpDst)
+		if err := f.Remove(tmpDst); err != nil {
+			return err
```
innobead, line 131: Should we handle this error?

innobead (changes requested):
In general, LGTM. Just a few comments. 

Just one feedback about when dealing an error but it's not critical and unhandled or it can be recovered eventually, suggest logging with warn level instead of error.

fsops/fsops.go
FrankYang0529: Yeah, I think we return error and for the caller to handle it. Or do you mean log it before return the error?

FrankYang0529:
> Just one feedback about when dealing an error but it's not critical and unhandled or it can be recovered eventually, suggest logging with warn level instead of error.

Thank you. I change the log level to warning.

fsops/fsops.go
innobead: In the experience, usually removing files would be best efforts, so the error could not be handled. Or, we can just have a warn log, good enough? (of course, those temp files could be leftover.)

PR #189, opened 2024-09-03
fix: refine lock related error message

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8713

use human readable info to replace lock type in the error message to make it clear to users.

### DIFF
--- a/lock.go
+++ b/lock.go
@@ -29,6 +29,12 @@ const BACKUP_LOCK LockType = 1
 const RESTORE_LOCK LockType = 1
 const DELETION_LOCK LockType = 2
 
+type Operation string
+
+const BackupOperationCreateRestore Operation = "create/restore"
+const BackupOperationDelete Operation = "delete"
+const BackupOperationUndefined Operation = "undefined"
+
 type FileLock struct {
 	Name       string
 	Type       LockType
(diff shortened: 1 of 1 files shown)


lock.go
```diff
@@ -29,6 +29,12 @@ const BACKUP_LOCK LockType = 1
 const RESTORE_LOCK LockType = 1
 const DELETION_LOCK LockType = 2
 
+type Operation string
+
+const CREATERESTORE Operation = "create/restore"
+const DELETE Operation = "delete"
+const UNDEFINED Operation = "undefined"
```
derekbit, line 36: The names are a bit ugly. Do we have better choices?
```suggestion
const (
  CREATERESTORE Operation = "create/restore"
  DELETE Operation = "delete"
  UNDEFINED Operation = "undefined"
)
```

derekbit: Why not separate CREATE and RESTORE

lock.go
```diff
@@ -107,16 +121,16 @@ func (lock *FileLock) Lock() error {
 	if !lock.canAcquire() {
 		file := getLockFilePath(lock.volume, lock.Name)
 		_ = removeLock(lock)
-		return fmt.Errorf("failed lock %v type %v acquisition", file, lock.Type)
+		return fmt.Errorf("failed to aquire lock %v when performing backup %v, please try again later.", file, operation)
 	}
 
 	file := getLockFilePath(lock.volume, lock.Name)
-	log.Infof("Acquired lock %v type %v on backupstore", file, lock.Type)
+	log.Infof("Acquired lock %v for backup %v on backupstore", file, operation)
```
derekbit, line 128: need `backup`?
```suggestion
	log.Infof("Acquired lock %v for %v on backupstore", file, operation)
```

lock.go
ChanYiLin: because lock type CREATE and RESTORE are the same, they are all "1" since they have the same priority.
Thus we can't distinguish what operation it is by only checking the lock

lock.go
ChanYiLin: The full message will be 
"Acquired lock /opt/backupstore/.../xxx.lck for **backup delete** on backupstore""
I feel like `backup delete` is better then only `delete`

lock.go
derekbit: I see. Thank you.

derekbit: Can we add a prefix to the constants?

ChanYiLin: Fixed. Thanks

derekbit (approved):
LGTM

PR #241, opened 2025-04-21
fix: update the progress if the backup already exists and we dont perform backup

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10740

There is a chance that the backup is triggered again even the content already exists in the backupstore
we should update the status to 100% so the CR status can have correct info


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #250, opened 2025-07-18
feat(backup): configurable bkup block size

COLDTURNIP:

Issue longhorn/longhorn#5215

Implement configurable backup block size.

- While creating a new backup, the block size is configured in the backup parameter. If the block size is not set, it fallback to default 2MB by default.
- The backup block size will be recorded into the backup configuration file.
- While restoring a backup, the block size is read from the backup configuration file.
- For incremental backups, the block size must be identical between backups.

This PR depends on longhorn/go-common-libs#122. The dependency commits must be resolved before merging.


backupstore.go
```diff
@@ -49,6 +49,7 @@ type Backup struct {
 	CompressionMethod     string
 	NewlyUploadedDataSize int64 `json:",string"`
 	ReUploadedDataSize    int64 `json:",string"`
+	BlockSize             int64 `json:",string,omitempty"`
```
derekbit, line 52: Need `omitempty`?

deltablock.go
```diff
@@ -562,11 +588,18 @@ func performBackup(bsDriver BackupStoreDriver, config *DeltaBackupConfig, delta
 	destURL := config.DestURL
 	concurrentLimit := config.ConcurrentLimit
 
+	blockSize, err := config.getBackupBlockSize()
+	if err != nil {
+		logrus.WithError(err).Errorf("Failed to get backup block size, use default size %v instead", DEFAULT_BLOCK_SIZE)
```
derekbit, line 593:
```suggestion
		logrus.WithError(err).Warnf("Failed to get backup block size, use default size %v instead", DEFAULT_BLOCK_SIZE)
```

derekbit: Do we really need the fallback?

COLDTURNIP: For the backup creation, we can return the error directly. All backup creation request should contain the block size parameter after v1.10.

backupstore.go
COLDTURNIP: No necessary. Let me remove it.

config.go
```diff
@@ -333,6 +333,10 @@ func loadBackup(bsDriver BackupStoreDriver, backupName, volumeName string) (*Bac
 		log.Infof("Fall back compression method to %v for backup %v", LEGACY_COMPRESSION_METHOD, backup.Name)
 		backup.CompressionMethod = LEGACY_COMPRESSION_METHOD
 	}
+	if backup.BlockSize == 0 {
+		log.Infof("Fall back block size to %v for backup %v", DEFAULT_BLOCK_SIZE, backup.Name)
```
derekbit, line 337:
```suggestion
		log.Infof("Falling back block size to %v for backup %v", DEFAULT_BLOCK_SIZE, backup.Name)
```

derekbit (dismissed):
LGTM

derekbit (approved):
LGTM

PR #268, opened 2026-01-13
feat(backup): support identity-based auth methods for azure

techknowlogick:

To support non-static tokens for Azure backup targets. As of right now, if you want to use Azure Blob storage as a backup target (unless you use a fragile workaround with k8s secrets), you need to use a static key.

This PR adds support for Azure managed identity, AKS workload identity and service principals.

If you're still keen on using static tokens but want to scope permissions, you can use service principals. Although this PR is backward compatible with the previous code/configuration, so you can keep using Account Key.

https://github.com/longhorn/longhorn/issues/12600

https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
https://learn.microsoft.com/en-us/azure/aks/use-managed-identity
https://learn.microsoft.com/en-us/azure/azure-arc/kubernetes/conceptual-workload-identity


derekbit:
@techknowlogick Thanks for your contribution. 
Could you create an IMPROVEMENT ticket in https://github.com/longhorn/longhorn/issues? It will help QA to verify the improvement. Thank you.

derekbit (commented):
In general, LGTM.

derekbit:
@techknowlogick Could you address the comments from @copilot? Thank you.

derekbit:
BTW @techknowlogick Could you create a PR to update https://github.com/longhorn/website/blob/master/content/docs/1.12.0/snapshots-and-backups/backup-and-restore/set-backup-target.md#set-up-azure-blob-storage-backupstore? Thanks.

derekbit:
@mantissahz @COLDTURNIP Please review the PR.

mantissahz:
In general, LGTM.
@techknowlogick, please fix the conflict. And could you explain why some packages are downgraded?
BTW, maybe we need to test the basic functionalities with the PR first.

PR #313, opened 2026-07-17
fix(s3): use recursive listing to avoid excessive API calls

0byte-coding:

Issue longhorn/longhorn#1547

`getBlockNamesForVolume` and `getAllBlockNames` walk the 2-level sharded block directory tree with one `driver.List()` call per directory, issuing up to 256*256 List calls per volume. On backends that bill per API request (e.g. Backblaze B2 Class C transactions), this is the main driver of the excessive call counts reported in longhorn/longhorn#1547 (one user measured up to 65,000 requests per backup).

This PR adds an optional `RecursiveLister` interface that a driver can implement to provide a native flat/recursive listing:
- `s3`: `ListRecursive` uses `ListObjectsV2` without a delimiter, returning every key under a prefix via normal pagination (1 API call per 1000 objects) instead of one call per directory.
- `fsops` (nfs/cifs/vfs): `ListRecursive` walks the local mount in a single process pass instead of one `List()` round-trip per directory level.

`getBlockNamesForVolume` / `getAllBlockNames` prefer the recursive path when the driver supports it, and fall back to the original walk-based implementation otherwise, so drivers that don't implement `RecursiveLister` (e.g. azblob) keep working unchanged.

#### Special notes for your reviewer:

**End-to-end verified**, not just unit tested. Built this patch into a full `longhorn` engine CLI binary (same binary `longhorn-manager` execs for `backup rm`/GC) against a real k3s + Longhorn v1.10.1 cluster and a real Backblaze B2 bucket with valid, working credentials. Ran `backup rm` on an identical 135-block backup with both the stock and patched binary, capturing every HTTP request via a local MITM proxy:

| | Stock binary | Patched binary | Change |
|---|---|---|---|
| Total S3 requests | 534 | 296 | -44.6% |
| Directory-listing GET calls | 243 | **5** | **-97.9%** |
| GC wall time | ~24s | ~13s | -46% |

Both runs produced identical, correct results (`Removed 135 unused blocks`, bucket left with only `volume.cfg`) - confirming no functional regression. The 5 remaining directory-listing calls (root check, `locks/`, `backups/` x2, `blocks/`) are legitimate; the `blocks/` listing that used to cost 243 nested calls is now a single `ListRecursive` call. The other ~137 GET calls in both runs are unrelated per-block existence checks (`FileSize`/`HeadObject`) done during GC bookkeeping, outside the scope of this fix.

Unit tests (`list_recursive_test.go`, `fsops/list_recursive_test.go`) cover the recursive listing path and the "directory doesn't exist" fallback case. Pre-existing test failures (`TestInspectBackup`, NFS mount tests requiring superuser) are unrelated to this change and reproduce identically on unmodified `master`.

See discussion in longhorn/longhorn#1547, particularly joshimoo's analysis of the `O(nodes) * O(vol * 4)` call cost and sedlund's independent proof-of-concept for the same recursive-listing approach.

### DIFF
--- a/backupbackingimage/config.go
+++ b/backupbackingimage/config.go
@@ -165,8 +165,36 @@ func GetAllBackupBackingImageNames(driver backupstore.BackupStoreDriver) ([]stri
 }
 
 func getAllBlockNames(driver backupstore.BackupStoreDriver) ([]string, error) {
-	names := []string{}
 	blockPathBase := getBackingImageBlockPath()
+
+	// See getBlockNamesForVolume in the top-level package for rationale:
+	// prefer a single recursive listing over walking the sharded
+	// 2-level block directory tree with one List() call per directory.
+	// https://github.com/longhorn/longhorn/issues/1547
+	if recDriver, ok := driver.(backupstore.RecursiveLister); ok {
+		// A missing directory is not an error: both the s3 and fsops
+		// ListRecursive implementations already return (nil, nil) for a
+		// prefix/path that doesn't exist, matching List()'s existing
+		// behavior below. Any error returned here is a genuine failure
+		// (auth, network, pagination, etc.) and must propagate to
+		// getBlockInfos/the caller rather than being silently treated as
+		// "no blocks" - otherwise backing-image deletion would report
+		// success while silently skipping block discovery and cleanup.
+		paths, err := recDriver.ListRecursive(blockPathBase)
+		if err != nil {
+			return nil, err
+		}
+		// ListRecursive returns paths relative to blockPathBase, e.g.
+		// "aa/bb/<checksum>.blk". ExtractNames/ValidateName only accept
+		// bare file names, so strip the shard directory components first.
+		baseNames := make([]string, len(paths))
+		for i, p := range paths {
+			baseNames[i] = filepath.Base(p)
+		}
+		return util.ExtractNames(baseNames, "", BlkSuffix), nil
+	}
+
+	names := []string{}
 	lv1Dirs, err := driver.List(blockPathBase)
 	// Directory doesn't exist
 	if err != nil {
--- a/cifs/cifs_test.go
+++ b/cifs/cifs_test.go
@@ -1,0 +1,19 @@
+package cifs
+
+import (
+	"testing"
+
+	"github.com/longhorn/backupstore"
+)
+
+// TestBackupStoreDriverDoesNotImplementRecursiveLister locks in that cifs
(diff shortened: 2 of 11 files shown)


christophersherman:
@0byte-coding This looks directly relevant to a current production failure mode also covered by longhorn/longhorn#8060: a volume with roughly 1.4 million block objects can spend longer than an hourly backup interval in retention GC while the deletion lock prevents the next backup from starting.

A source audit confirms that current backupstore still walks the two-level block shard tree serially before serial per-block cleanup, so this recursive-listing change is the right first optimization and avoids changing deletion semantics. The DCO check is currently ACTION_REQUIRED; could you sign off and update the commit?

@mantissahz Since #8060 is assigned to you and marked priority/0, could you help route review once DCO is green? Per-block removal may merit a separately benchmarked bounded-concurrency follow-up, but #313 already removes the largest fixed LIST amplification.

0byte-coding:
Thanks for the confirmation and the pointer to #8060 - that lines up with what I found.

Force-pushed with:
- DCO sign-off (`git commit -s`) - now green.
- Addressed all 4 Copilot review comments:
  1. `list_recursive_test.go`: made shard directory creation explicit instead of relying on `afero.MemMapFs` auto-vivifying parent dirs (an implementation detail of that specific in-memory fake, not something `afero.WriteFile` itself guarantees).
  2. `deltablock.go`: `getBlockNamesForVolume` now propagates genuine `ListRecursive` errors instead of treating every error as "directory doesn't exist" - both the s3 and fsops implementations already return `(nil, nil)` for a missing prefix/path, so any error is a real failure (auth/network/pagination) that should abort GC rather than silently persisting `BlockCount = 0`.
  3. `backupbackingimage/config.go`: same fix applied to `getAllBlockNames`.
  4. `fsops/fsops.go`: renamed the walk-based helper to `ListRecursiveLocal` and stopped it being promoted automatically via embedding. Only `vfs` (a plain local directory) now opts into `RecursiveLister` explicitly; `nfs`/`cifs` (network mounts) deliberately do not, since `filepath.Walk` still issues one remote metadata round-trip per directory entry there - it could turn one round-trip per directory level into one round-trip per block on a volume with a large number of blocks. Added compile-time interface-assertion tests locking this in for all three drivers.

Also added a regression test for point 2 (`TestGetBlockNamesForVolumePropagatesRecursiveListError`).

Agreed per-block removal in GC (the DELETE side) looks like a good separately-benchmarked follow-up - this PR only touches the LIST side.

PR #314, opened 2026-07-21
fix(s3): avoid multipart upload for backup metadata and block writes

derekbit:

Issue Longhorn/longhorn#13573

BackupStoreDriver.Write() uploads backup_*.cfg, volume.cfg, and individual data blocks via service.PutObject(), which uses manager.Uploader. That uploader defaults PartSize to 5 MiB (manager.DefaultUploadPartSize): once the payload exceeds it, the SDK silently switches from a single PutObject request to a multipart upload (CreateMultipartUpload + UploadPart + CompleteMultipartUpload).

backup_*.cfg grows with the number of blocks in a backup (~109 bytes per block entry), so it crosses the 5 MiB threshold on large volumes. Some S3-interop providers (e.g. Google Cloud Storage) reject the resulting multipart request shape with:
"AWS Error: SignatureDoesNotMatch Invalid argument" even though a plain PutObject request of the same object succeeds. This reproduces 100% of the time on volumes with enough blocks to push backup_*.cfg past 5 MiB, while smaller volumes (and all data blocks, which stay well under 5 MiB) are unaffected.

Add service.PutObjectAsSinglePart(), which always issues a single PutObject request regardless of payload size, bypassing manager.Uploader entirely. Route BackupStoreDriver.Write() through it, since all of its payloads (metadata and individual blocks) are well under the 5 GiB limit of a single PutObject request. Fail fast with a clear error if a payload ever exceeds that limit, instead of letting the request reach S3 and come back as an opaque EntityTooLarge error.

BackupStoreDriver.Upload() (single-file backups, system-backup bundles) is unchanged and still uses manager.Uploader, keeping multipart upload's parallelism for genuinely large files.

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -180,7 +180,15 @@ func (s *BackupStoreDriver) Read(src string) (io.ReadCloser, error) {
 
 func (s *BackupStoreDriver) Write(dst string, rs io.ReadSeeker) error {
 	path := s.updatePath(dst)
-	return s.service.PutObject(context.Background(), path, rs)
+	// Driver.Write only carries backup metadata (backup_*.cfg, volume.cfg) and
+	// individual data blocks, all well under the 5 GiB single-PutObject limit.
+	// Force a single PutObject request here to avoid the manager.Uploader's
+	// default 5 MiB PartSize silently switching metadata writes to a multipart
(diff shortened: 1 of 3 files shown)


mantissahz (approved):
In general, LGTM

